Issue #1, opened 2016-08-18
Add support for second passphrases


karmacoma:
Accounts with a second passphrase enabled are unable to send LSK. We should add support for second passphrases so this is possible.

fix:
There is secondsignature swicth in the json you get when polling an account, see http://108.61.99.136:8000/api/accounts?address=7136766316744460715L

Then using lisk js, is just adding the second passphrase as last arg when creating transaction

PR #4, opened 2016-08-31
Fix to #1

fix:
This is a simple fix (not tested) to #1


rferro:
@fix I fix this with a modal to the user enters the second passphrase after clicking to send. I think that it was better. Ty

fix:
yes sounds even better to me, nice job!

PR #35, opened 2017-02-27
Ctrl+a to select passphrase on registration - Closes #15

### LINKED ISSUE #15: Ctrl/Cmd + A select more than passphrase
When user creates new account, "select all" - ctrl/cmd+a selects entire website behind and passphrase instead of just passphrase.

screen shot 2016-10-27 at 11 57 54

slaweet:
I'm not entirely happy about setting a fixed width to the textarea, but otherwise it just is really narrow.

Closes #15


4miners:
Did you test that on mobile?

PR #39, opened 2017-03-02
Fix multiple new passphrase windows bug - Closes #16

### LINKED ISSUE #16: New passphrase window sometimes pops-up more than once
screen shot 2016-10-27 at 12 00 39

Hard to reproduce, just sometimes pop up twice, sometimes it removes previous one, sometimes there are 2 windows stacked.

slaweet:
Removing $timeout seems to fix the issue. I was able to reproduce it multiple times, before doing the change and never after doing the change. I'm just worried what was a purpose of the $timeout being there in the first place.

Closes #16

### DIFF
--- a/src/app/components/login/login.jade
+++ b/src/app/components/login/login.jade
@@ -3,10 +3,10 @@ md-card
     md-card-title-text
       span.md-title Sign In
   md-card-content
-    form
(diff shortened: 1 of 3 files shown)


slaweet:
Sorry about adding other commits to this pull request, they didn't like big enough changes to have their own branch.

Issue #80, opened 2017-03-29
Floating-point arithmetic rounding error in balance


slaweet:
balance returned form the API is: `"balance":"9999997590000000"`
but the ui shows: 99999975.90000001 LSK
![round_issue](https://cloud.githubusercontent.com/assets/1254342/24456643/f80a95ae-1493-11e7-99d8-8063dc403b15.png)

slaweet:
I think it happens only for an account with very high balance, thus it won't happen to any user on main net. I tried to reproduce it with lower balance, but it didn't happen.

karmacoma:
Closed by #90

PR #90, opened 2017-04-04
Fix floating-point arithmetic rounding error in balance #80

slaweet:
The error was:
balance returned form the API is: "balance":"9999997590000000"
but the ui shows: 99999975.90000001 LSK

Fixed by switching from numeral to BigNumber.js

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -20,13 +20,13 @@
     "angular-material-data-table": "=0.10.9",
     "angular-messages": "=1.5.8",
     "babel-polyfill": "=6.9.1",
+    "bignumber.js": "^4.0.0",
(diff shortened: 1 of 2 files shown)


src/package.json
```diff
@@ -20,13 +20,13 @@
     "angular-material-data-table": "=0.10.9",
     "angular-messages": "=1.5.8",
     "babel-polyfill": "=6.9.1",
+    "bignumber.js": "^4.0.0",
```
karmacoma: Please fixate to `4.0.0`.

slaweet: Sure.
I was wondering if there is a npm setup to do it for me..
And it turns out there is:
```
npm config set save-prefix '='
```

PR #102, opened 2017-04-06
Move network selection to login module - Closes #89

### LINKED ISSUE #89: Select network in login dialog
It doesn't really make sense change network after login, as user most likely doesn't have the same account address on all mainnet, testnet, localhost.
Just add a simple dropdown with 3 options:
 - Official, 
 - Testnet, 
 - Localhost

The first should be the default.

Then the peer card in the UI after login should only list nodes on the selected network.

alihaghighatkhah:
Changes made:
 - The network select element and corresponding watcher is moved to login module
 - Instead added a label showing the selected network in top module
 - Tests are changed accordingly

Closes #89

### DIFF
--- a/src/app/components/login/login.pug
+++ b/src/app/components/login/login.pug
@@ -2,10 +2,15 @@ md-card
   md-card-title
     md-card-title-text
       span.md-title Sign In
-  md-card-content
+  md-card-content(flex='100', flex-gt-sm='70', flex-offset-gt-sm='15')
     form(ng-submit='$ctrl.doTheLogin()')
       md-input-container.md-block(md-is-error='$ctrl.valid === 0')
-        label Enter your passphrase
+        label.select Choose your network
+        md-select(ng-model='$ctrl.$peers.currentPeerConfig', aria-label='Peer')
+          md-optgroup(ng-repeat='(name, peers) in $ctrl.$peers.stack', ng-if='peers.length', label='{{ name }}')
+            md-option(ng-repeat='peer in peers', ng-value='peer') {{ peer.node }}{{ peer.port ? ':' + peer.port : '' }}
+      md-input-container.md-block(md-is-error='$ctrl.valid === 0')
+        label.pass Enter your passphrase
         input(type="{{ $ctrl.show_passphrase ? 'text' : 'password' }}", ng-model='$ctrl.input_passphrase', ng-disabled='$ctrl.generatingNewPassphrase', autofocus)
       md-input-container.md-block
         md-checkbox.md-primary(ng-model="$ctrl.show_passphrase", aria-label="Show passphrase") Show passphrase
--- a/src/app/components/login/login.pug
+++ b/src/app/components/login/login.pug
@@ -5,7 +5,7 @@ md-card
   md-card-content(flex='100', flex-gt-sm='70', flex-offset-gt-sm='15')
     form(ng-submit='$ctrl.doTheLogin()')
       md-input-container.md-block(md-is-error='$ctrl.valid === 0')
-        label.select Choose your network node
+        label.select Choose a peer
         md-select(ng-model='$ctrl.$peers.currentPeerConfig', aria-label='Peer')
           md-optgroup(ng-repeat='(name, peers) in $ctrl.$peers.stack', ng-if='peers.length', label='{{ name }}')
             md-option(ng-repeat='peer in peers', ng-value='peer') {{ peer.node }}{{ peer.port ? ':' + peer.port : '' }}
--- a/src/app/components/login/login.pug
+++ b/src/app/components/login/login.pug
@@ -5,7 +5,7 @@ md-card
   md-card-content(flex='100', flex-gt-sm='70', flex-offset-gt-sm='15')
     form(ng-submit='$ctrl.doTheLogin()')
       md-input-container.md-block(md-is-error='$ctrl.valid === 0')
-        label.select Choose your network
+        label.select Choose your network node
         md-select(ng-model='$ctrl.$peers.currentPeerConfig', aria-label='Peer')
           md-optgroup(ng-repeat='(name, peers) in $ctrl.$peers.stack', ng-if='peers.length', label='{{ name }}')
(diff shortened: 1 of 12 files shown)


src/karma.conf.js
```diff
@@ -95,6 +95,7 @@ module.exports = function (config) {
     // If true, Karma captures browsers, runs the tests and exits
     singleRun: !opts.live,
     client: {
+      captureConsole: true,
```
slaweet: This should go a separate commit

src/webpack.config.babel.js
```diff
@@ -16,6 +16,7 @@ const PATHS = {
 };
 
 const common = {
+  devtool: 'source-map',
```
slaweet: This should go a separate commit

src/spec/spec.js
```diff
@@ -120,22 +125,19 @@ function testAddress() {
 }
 
 function testPeer() {
-  login(masterAccount);
-  waitForElemAndCheckItsText('.peer md-select-value .md-text', 'localhost:4000');
+  expect(element.all(by.css('form md-select md-select-value span:first-child')).get(0).getText()).toEqual('Choose your network node');
```
slaweet: This test doesn't work if you run only this one because you first need to `launchApp();`

src/app/components/top/top.pug
```diff
@@ -9,9 +9,7 @@ md-content(layout='column', layout-gt-xs='row')
         i.material-icons.offline(ng-show='!$ctrl.$peers.online') error
         i.material-icons.online(ng-show='$ctrl.$peers.online') check circle
       span.md-title.title Peer
-      md-select(ng-model='$ctrl.$peers.currentPeerConfig', aria-label='Peer')
-        md-optgroup(ng-repeat='(name, peers) in $ctrl.$peers.stack', ng-if='peers.length', label='{{ name }}')
-          md-option(ng-repeat='peer in peers', ng-value='peer') {{ peer.node }}{{ peer.port ? ':' + peer.port : '' }}
+      div.address.value {{ $ctrl.$peers.currentPeerConfig.node }}:{{ $ctrl.$peers.currentPeerConfig.port }}
```
slaweet: .address class shouldn't be here, it causes this warning in e2e tests:
```
W/element - more than one element found for locator By(css selector, .address) - the first result will be used
```

karmacoma:
> Login screen layout is broken in Safari, but I'm not sure if we care

I think we should try to fix at least. Or at least find out why it is broken specifically on Safari.

src/karma.conf.js
karmacoma: Agree

src/package.json
```diff
@@ -52,7 +52,7 @@
     "jasmine-spec-reporter": "^3.2.0",
     "jit-grunt": "=0.10.0",
     "json-loader": "=0.5.4",
-    "karma": "^1.5.0",
+    "karma": "1.4.1",
```
karmacoma: `=1.4.1`

src/spec/spec.js
```diff
@@ -83,11 +84,15 @@ function testNewAccount() {
   launchApp();
 
   element.all(by.css('.md-button.md-primary')).get(0).click();
+  /**
+   * To generate a random pattern of mousemove events, we're randomizing the x,y pairs that
+   * based on even/odd valkuds of i will locate right/left halves of screen
```
karmacoma: Spelling and grammar.

```
/**
 * To generate a random pattern of mousemove events, we are randomizing the x,y pairs that
 * based on even/odd values we locate the right/left halves of the screen
 */
```

src/test/components/login/login.spec.js
```diff
@@ -67,6 +72,29 @@ describe('Login controller', () => {
     controller.passphrase = '';
   });
 
+  describe('controller()', () => {
+    it('Should define a watcher for $ctrl.$peers.currentPeerConfig', () => {
```
karmacoma: `should`

src/test/components/login/login.spec.js
```diff
@@ -67,6 +72,29 @@ describe('Login controller', () => {
...
+      const peers = controller.$peers;
+      const spy = sinon.spy(peers, 'setActive');
+      peers.currentPeerConfig = peers.stack.localhost[0];
+      $scope.$apply();
+      peers.currentPeerConfig = peers.stack.official[0];
+      $scope.$apply();
+      expect(spy).to.have.been.calledWith();
+    });
+
+    it('Should be able to change the active peer', () => {
```
karmacoma: `should`

src/webpack.config.babel.js
karmacoma: Agree

src/app/components/login/login.pug
```diff
@@ -2,10 +2,15 @@ md-card
   md-card-title
     md-card-title-text
       span.md-title Sign In
-  md-card-content
+  md-card-content(flex='100', flex-gt-sm='70', flex-offset-gt-sm='15')
     form(ng-submit='$ctrl.doTheLogin()')
       md-input-container.md-block(md-is-error='$ctrl.valid === 0')
-        label Enter your passphrase
+        label.select Choose your network node
```
karmacoma: I would phrase it as: `Choose a peer`.

Also, upon logging in, the selection is referred to as a "Peer", not a node.

I'm happy to see either or, as long as the terminology is consistently applied.

src/package.json
alihaghighatkhah: 1.5.0 has some bugs, according to the authors

alihaghighatkhah:
I'm not sure if the interpolation issue with the active peer is related to this ticket. based on the templates I suppose such issue must happen in many cases. besides, we're using interpolation markup in many cases, taking in to account that it uses watchers, replacing it with ngBind, ngValue and other directives can help improve performance.

karmacoma:
@alihaghighatkhah Can we remove colon from hostname when there is no port? Looks like so atm:

image

karmacoma:
@Tosch110 @alihaghighatkhah: If there is a failure to connect to a selected peer, then the current behaviour results in a another mainnet node being selected. Even if localhost or tesnet peer is chosen. I think this should be fixed in: `lisk-js`. We should keep users on same network at least.

PR #172, opened 2017-05-03
Fixing appearance issues in Firefox - Fixes #147

### LINKED ISSUE #147: Appearance issues in Firefox
### Expected appearance

 - Send modal expands to correct height, not overlapping footer.
 - The correct tab should be activated in main page first time logging in.
 - The underlay of the send modal must expand to cover entire page.
 - Show scrollbars for whole body.

### Actual appearance

image
 - Always the first tab is activated when logging in.
 - The dark underlay in not expanding the whole page in for send modal.
 - scrollbars are showing for the main container instead of whole body.

### Steps to reproduce
Using Mozilla Firefox `53.0`:
  
- About the modal appearance:
  - Log into `lisk-nano`.
  - Click `Send`.
- About the tabs
  - Activate voting tab
  - Log out
  - Log in
  - Now the first tab is activated instead of the second tab.
- About the send modal:
  - Click on Send button
  - Now the dark underlay is only covering the main template instead of the whole page.
- About scrollbars
  - Go to voting tab
  - scroll down
  - Now scrollbars are attached to the vote list instead of the edge of the window

alihaghighatkhah:
- Fix an issue with body overflow
 - Activate the correct tab when initialising the application
 - Expand modal underlaying layer to entire view
 - Hide modals before the state changes

Fixes #147

### DIFF
--- a/src/app/components/delegates/vote.js
+++ b/src/app/components/delegates/vote.js
@@ -18,7 +18,7 @@ app.component('vote', {
       this.votedList = [];
 
       this.delegateService.listAccountDelegates({
-        address: this.account.address,
+        address: this.account.get().address,
       }).then((data) => {
         this.votedList = data.delegates || [];
         this.votedList.forEach((delegate) => {
--- a/src/app/components/main/main.js
+++ b/src/app/components/main/main.js
@@ -71,7 +71,7 @@ app.component('main', {
           this.account.set({ balance: res.balance });
         })
         .catch((res) => {
-          this.account.get({ balance: null });
+          this.account.set({ balance: null });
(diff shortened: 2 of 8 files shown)


slaweet:
The test that is failing here is unstable:
 ✖ calls this.$peers.active.getAccountPromise(this.address) and if it fails, then resets this.account.balance and reject the promise that update() returns
      expected 0 to equal undefined

I was fixing the issue in one of my own PRs with this commit:
https://github.com/LiskHQ/lisk-nano/pull/165/commits/99f81e08f53c23cd190a09d2150eeac09f0fb8ac

PR #177, opened 2017-05-04
Change the style of Send modal to make it similar to other modals - Closes #175

### LINKED ISSUE #175: Change the appearance of the Send modal similar to Vote modal
### Expected appearance
We want all of our modal use an identical design. Presently all the modals have blue headers and expanded form elements. Send Modal should be the same.

### Actual appearance
 - The header is white.
 - The form is not expanded to entire width of the modal.
 - Action buttons are different from those in other modals.

alihaghighatkhah:
This is using the same styling as other modals are.

Closes #175


karmacoma:
@alihaghighatkhah: After merging latest development, the following test is failing.

```
 ✖ 1 test failed

FAILED TESTS:

  Forging component

    ✖ "before each" hook for "should contain a card with delegate name"

      PhantomJS 2.1.1 (Linux 0.0.0)

    undefined is not an object (evaluating 'account.network')

    setActive

    

    $emit
```

PR #188, opened 2017-05-08
Fix routing issues - Closes #180

### LINKED ISSUE #180: http://localhost:8080/ is broken
### Expected behaviour
http://localhost:8080/ shows the app
### Actual behaviour
http://localhost:8080/ redirects to http://localhost:8080/#/404 which displays just the LIsk Nano logo.
The url that works is http://localhost:8080/#/, so http://localhost:8080/ should redirect to that
### Steps to reproduce
- open browser
- go to http://localhost:8080/

alihaghighatkhah:
- Redirect unknown routes to login page.
- Set the transactions tab as default tab of the main page.
- Redirect to landing page from login instead of main.

### DIFF
--- a/src/app/components/login/login.js
+++ b/src/app/components/login/login.js
@@ -77,9 +77,6 @@ app.component('login', {
       const passphrase = this.$location.search().passphrase || this.$cookies.get('passphrase');
       if (passphrase) {
         this.input_passphrase = passphrase;
-        if (this.$rootScope.logged === undefined) {
-          // this.$timeout(this.passConfirmSubmit.bind(this), 10);
-        }
       }
     }
   },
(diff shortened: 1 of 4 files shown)


src/app/components/login/login.js
```diff
@@ -79,7 +78,7 @@ app.component('login', {
       if (passphrase) {
         this.input_passphrase = passphrase;
         if (this.$rootScope.logged === undefined) {
-          this.$timeout(this.passConfirmSubmit.bind(this), 10);
+          // this.$timeout(this.passConfirmSubmit.bind(this), 10);
         }
```
karmacoma: Why not remove whole `if` branch?

slaweet: I believe it's for development purposes. Just uncomment it and it does autologin.

slaweet: I guess it should be described in a comment in the file

alihaghighatkhah: Removed it

PR #204, opened 2017-05-10
Fix delegate registration for account with second passphrase - Closes #198

### LINKED ISSUE #198: Account with second passphrase cannot register as delegate
### Expected behaviour
"Register as delegate" dialog should include "2nd passphrase" field if the account has it.
### Actual behaviour
registre_2nd

slaweet:
Closes #198


Isabello:
I reran this with the change from #205

PR #210, opened 2017-05-11
Eliminate redundant Api calls. Closes #174

### LINKED ISSUE #174: Use window.requestAnimationFrame() instead of timeouts
### Actual behaviour
We are setting timeouts (usually 20 seconds) in various components to re-check for things like:
- latest transactions
- account balance
- delegates
- votes

### Expected behaviour
As @alihaghighatkhah pointed out, we should be using `window.requestAnimationFrame()` because it triggers only when the window is visible, so it can save a lot of traffic.
Although, we must implement a function that determines whether it makes sense to repeat a particular request. It would be nice if it could somehow synchronise with block time so that it knows that a new block has been forged and thus there can potentially be some changes. 

Maybe we can use this angular wrapper:
https://www.npmjs.com/package/angular-raf

alihaghighatkhah:
- Remove recursive timeouts in all the components
 - Create a Sync service to manage intervals 
 - Deep compare the result of status check and broadcast `accountChange` only if values are actually changed
 - Use camelCase for event names


src/app/components/main/main.js
```diff
@@ -32,15 +30,23 @@ app.component('main', {
 
       this.$rootScope.prelogged = true;
 
-      this.update()
+      this.update(attempts)
         .then(() => {
           this.$rootScope.prelogged = false;
           this.$rootScope.logged = true;
-          this.checkIfIsDelegate();
+          if (this.$timeout) {
```
slaweet: I don't understand the purpose of this `if` statement, but it looks weird. Maybe add some comment.

src/app/services/account.js
```diff
@@ -3,33 +3,63 @@ import lisk from 'lisk-js';
...
+    /* eslint-disable eqeqeq */
+
+    if (ref1 == undefined && ref2 == undefined) {
+      return true;
+    }
+    if (typeof ref1 !== typeof ref2 || (typeof ref1 !== 'object' && ref1 != ref2)) {
+      return false;
+    }
+
+    const props1 = (ref1 instanceof Array) ? ref1.map((val, idx) => idx) : Object.keys(ref1).sort();
```
alihaghighatkhah: What this function does is to make sure we're updating list only when a real change is happening. so it doesn't perform a crucial task. though to meet 90+% it makes sense to cover it.

PR #224, opened 2017-05-15
Implement try/catch into Jenkinsfile - Closes #223

### LINKED ISSUE #223: Implement Try/Catch in Jenkins File
### Expected behaviour
Jenkins should throw failures and stop the build.

### Actual behaviour
Jenkins will continue to the end of the build regardless of failure status

### Steps to reproduce
Run a build that fails on purpose.

Isabello:
Best practice for jenkins is to include try/catch clauses for each stage so that the build can be errored out appropriately if theres a failure. This PR implements just that.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -81,6 +81,9 @@ node('master-nano-01'){
     stage ('Start Dev Server and Run Tests') {
       try {
         sh '''
+        # Prepare lisk core for testing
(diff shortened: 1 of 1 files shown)


Isabello:
Passphrase generation is still fragile for some reason. Might be worth a second look.

Isabello:
retest this please

PR #231, opened 2017-05-16
Rename `Transfer` -> `Send`

### LINKED ISSUE #215: Revert "Transfer" back to "Send"
### Expected appearance

Button and window title originally used the verb "Send".

### Actual appearance

Button and window title now uses the verb "Transfer".

image

image

### Steps to reproduce

Launch the development version of `1.0.0`.

alepop:
Resolve #215

### DIFF
--- a/src/app/components/header/header.pug
+++ b/src/app/components/header/header.pug
@@ -1,7 +1,7 @@
 md-content.header(layout='row', layout-align='center center', layout-padding)
   img.logo(src=require('../../assets/images/LISK-nano.png'))
   div(flex)
-  md-button.md-raised.md-primary.transfer(data-show-transfer-modal, ng-if='$root.logged') Transfer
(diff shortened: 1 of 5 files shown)


alihaghighatkhah:
To the best of my knowledge in financial terminology the verb _Transfer_ is preferred to _Send_.
As you can see [here](https://en.wikipedia.org/wiki/Transfer_payment), [here](https://en.wikipedia.org/wiki/Balance_transfer), [here](https://en.wikipedia.org/wiki/Money_transfer) and many other related documentation.
So I would suggest discarding this PR and closing the ticket.

Isabello:
I agree, we should always be using the word transfer rather than send. We are always transfering tokens from one place to another, not sending them.

alihaghighatkhah:
Since we haven't concluded on changing the wording and in case of changing, all the occurrence of word _Transfer_ in templates and scripts should change to maintain the consistency, I close this PR.

PR #237, opened 2017-05-17
Fix forging tab visibility - Closes #218

### LINKED ISSUE #218: Forging tab should be always visible
### Expected behaviour

User should be able to click on a visible "Forging" tab and then: "Register as a delegate".

### Actual behaviour

No forging tab is displayed.

Secondly, by navigating to the route `/#/main/forging` for a non-delegate, the user is presented with a disabled button: "Become Delegate". Therefore a dead end.

image

### Steps to reproduce

Log in as a non-delegate.

slaweet:


src/app/components/delegate-registration/delegateRegistration.js
```diff
@@ -1,28 +1,39 @@
...
+    function checkPendingRegistration() {
+      delegateService.getDelegate({
+        username: $scope.username,
+      }).then((data) => {
+        Account.set({
+          isDelegate: true,
+          username: data.delegate.username,
+        });
+      }).catch(() => {
+        $timeout(checkPendingRegistration, 10000);
```
alihaghighatkhah: Please use Sync service instead of timeout

src/app/components/login/login.js
```diff
@@ -57,7 +57,9 @@ app.component('login', {
           passphrase: this.Passphrase.normalize(_passphrase),
           network: this.network,
         });
-        this.$state.go(this.$rootScope.landingUrl || 'main.transactions');
+        this.$state.go(this.$rootScope.landingUrl && this.$rootScope.landingUrl !== 'main.forging' ?
```
alihaghighatkhah: Please remove this part and make sure the message in `src/app/components/forging/forging.pug` will be shown if the user is not delegate or the registration is not confirmed,

src/app/components/main/main.js
```diff
@@ -65,6 +65,9 @@ app.component('main', {
             isDelegate: true,
             username: data.delegate.username,
           });
+          if (this.$rootScope.landingUrl === 'main.forging') {
```
alihaghighatkhah: Same for here

PR #263, opened 2017-05-24
Show desktop notifications for changes

### LINKED ISSUE #253: Show desktop notifications for changes
### Expected behaviour
Would be cool if we can show notifications when the account's balance has been changed. I mean when a deposit happens

### Actual behaviour
Presently we don't have any notification for this kind of changes.

### Steps to reproduce
- Send some funds to your account.
- You won't get any updates unless with opening the application and reading the balance.

alepop:
Resolve #253


slaweet:
@alepop I tried to test this PR, but I couldn't get it to show any notification. I think it's related to the fact that we're using `window.requestAnimationFrame` (in sync service [0]) to get account updates, so it doesn't check server when the app window is not active.

[0] https://github.com/LiskHQ/lisk-nano/blob/development/src/services/sync.js

alepop:
@slaweet I fix this issue in commit 7677d26
```
+    webPreferences: {
 +      // Avoid app throttling when Electron is in background
 +      backgroundThrottling: false,
 +    },
```

slaweet:
@alepop thanks for the update. Now that I knew it should be working I found out that the notification was appearing on a different screen than the window was. But I suppose that's just my own settings.
So the feature works. 
Still, I'm not sure if this feature is worth all the requests that otherwise could have been spared while the app is in the background. @alihaghighatkhah , @karmacoma what do you think?
Maybe we can decrease the request interval when in background.

alihaghighatkhah:
@slaweet  Decreasing the sync broadcast rate could be a solution. I don't have anything against it.

alepop:
@slaweet, @alihaghighatkhah  what do you think the best interval in the inactive window `20000`, `30000` ?

slaweet:
I would go even with `60000`. 
In future, lisk core should provide some web socket API for this use case, so we don't have to choose arbitrary poll intervals.

src/components/main/main.js
```diff
@@ -17,7 +17,7 @@ app.component('main', {
    */
   controller: class main {
     constructor($scope, $rootScope, $timeout, $q, $state, Peers,
-      dialog, SendModal, Account, AccountApi) {
+      dialog, SendModal, Account, AccountApi, Notify) {
```
alihaghighatkhah: Please change the name of the service to `Notification`.

src/components/main/main.js
```diff
@@ -110,6 +111,10 @@ app.component('main', {
             delete res.publicKey;
           }
 
+          this.notify.about(
+            'deposite',
```
alihaghighatkhah: Please correct the typo: deposite -> deposit. Please make sure it's fixed everywhere.

src/components/main/main.js
```diff
@@ -110,6 +111,10 @@ app.component('main', {
             delete res.publicKey;
           }
 
+          this.notify.about(
```
alihaghighatkhah: Please check if `res.balance > this.account.get().balance` then call this method.

src/services/notify.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @description This factory provides methods to call Notification
+ *
+ * @module app
+ * @submodule Notify
+ */
+app.factory('Notify', ($window, lsk) => {
```
alihaghighatkhah: As mentioned before, Please change the name of the factory to `Notification`;

src/services/notify.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Creating notification about deposit
+     *
+     * @param {number} amount
+     * @private
+     * @memberof Notify
+     */
+    __deposite(amount) { // eslint-disable-line
```
alihaghighatkhah: Please keep only one underscore sign and fix the typo. The method name must be `_deposit`.

src/services/notify.js
```diff
@@ -0,0 +1,73 @@
...
+      if (amount > 0) {
+        new $window.Notification( // eslint-disable-line
+          'LSK received',
```
alihaghighatkhah: Please remove the new line. simply use white space between the string and the brace.

src/index.pug
```diff
@@ -7,6 +7,10 @@ html
       body {
         background-color: #eee !important;
       }
+    script.
```
alihaghighatkhah: Webpack has a built-in plugin named `ProvidePlugin`, you can use to create global variables. and since we're just using this for electron app, we can check the environment and add it only when building production package.

reyraa (changes requested):

src/services/notify.js
alihaghighatkhah: and since you're checking while calling this function in `main`, there's not need to check here any more.

src/components/main/main.js
alepop: I use subtraction because Notification display how many `LSK` received, while `res.balance` and `this.account.get().balance` display total balance amount.

alihaghighatkhah: Yes, exactly. So you check the the balance is increased, then you pass the increased amount to `Notification.about`. My point is we should check and call if necessary, not always call and then perform if necessary.

alepop: I understood you

src/index.pug
alepop: Maybe are you talking about `DefinePlugin`? Because `ProvidePlugin` will load the module but not create a global variable.

alihaghighatkhah: To the best of my knowledge `ProvidePlugin` can also do that. but sure, use which ever you find more fit. I just mean not to use script tags.

alepop: I came to the conclusion that the use `window.require` in script tag is the simplest way to loading `electron`. Because of `DefinePlugin` use common `require` and it does not load electron to the browser.

reyraa (dismissed):

slaweet (dismissed):

reyraa (approved):

reyraa:
Thank you @alepop

PR #268, opened 2017-05-29
Put delegates list to md virtual repeat - Closes #260

### LINKED ISSUE #260: Delegate list gets slow when loading ~1000 delegates
### Expected behaviour
The voting UI should be always smooth.
### Actual behaviour
The voting UI is lagging if many delegates are displayed.
### Steps to reproduce
- Open Lisk Nano
- Log in to an account on mainnet
- Go to "Voting" tab
- Scroll down like crazy till all delegates are loaded
- Attempt to select some of them

slaweet:
in order to speed up the ui when the list list long


src/components/delegates/delegates.js
```diff
@@ -41,11 +42,12 @@ app.component('delegates', {
       this.loading = true;
       this.usernameInput = '';
       this.usernameSeparator = '\n';
+      this.delegates = [];
 
       this.updateAll();
 
       this.$scope.$watch('search', (search, oldValue) => {
-        this.delegatesDisplayedCount = 20;
+        this.delegates = [];
```
alihaghighatkhah: With this solution we're facing 2 issues:
 - with every change on search phrase we're performing the heavy task of compiling the whole lisk. as we know angular's Achilles' heel. so we end up a glitchy behaviour.
 - Some times the doesn't fill the screen and you only can scroll in a small chunk of the screen whilst the rest of it is empty.
screen shot 2017-05-30 at 12 02 25

src/components/delegates/delegates.js
```diff
@@ -54,6 +56,39 @@ app.component('delegates', {
       this.$scope.$on('peerUpdate', () => {
         this.updateAll();
       });
+
+      const $ctrl = this;
```
alihaghighatkhah: This part is inconsistent with the rest of the code which uses `this`. may be using `function.bind` solves this issue.

src/components/delegates/delegates.js
```diff
@@ -54,6 +56,39 @@ app.component('delegates', {
...
+     * @method setVirtualRepeatContainerHeight
+     */
+    setVirtualRepeatContainerHeight() {
+      const setHeight = () => {
+        const element = angular.element('delegates md-virtual-repeat-container');
+        const repeatSizerHeight = angular.element('.md-virtual-repeat-sizer').height();
+        const newHeight = Math.round(Math.min(
+          repeatSizerHeight + 30,
+          this.$window.innerHeight * 0.7));
+        angular.element(element).css('height', `${newHeight}px`);
```
alihaghighatkhah: The default unit in `css` method is `px`, so you don't need this string template.

slaweet:
After a discussion with @alihaghighatkhah I'm closing this PR, because the virtual repeat causes too much trouble. Instead, we should migrate to Angular 2 or ReactJs or some other up-to-date framework with better performance

PR #271, opened 2017-05-29
Send from address or amount - Closes #270

### LINKED ISSUE #270: Send funds by clicking on account balance and transaction amount
### Expected behaviour
Presently it's not possible to 
 - click on account balance and send all funds.
 - click on sender/recipient address and send to that account (although there is a separated button for that purpose)
 - click on the amount of each transaction item and create a transaction with the same amount

### Actual behaviour
Would be great to have the above mentioned possibilities and add an appropriate  tooltip for each case.

### Steps to reproduce
 - Login and activate transactions tab. 
 - Now clicking on sender/recipient address or account balance won't open a modal to send funds

alihaghighatkhah:
Provides the functionality of showing send dialog with pre-filled values of address and/or amount by clicking on address or amount in transaction items and also by clicking on account balance.


src/components/transactions/transactions.pug
```diff
@@ -32,18 +31,20 @@ md-card.offline-hide
...
                 i.material-icons.in(ng-if='transaction.senderId !== $ctrl.account.get().address') call_received
                 i.material-icons.out(ng-if='transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address') call_made
               td(md-cell)
-                .amount(ng-class='{ neutral:transaction.type === 0 &&  transaction.senderId === transaction.recipientId, positive: transaction.senderId !== $ctrl.account.get().address, negative: transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address }')
+                .amount(ng-class='{"has-send-modal": transaction.amount > 0, neutral:transaction.type === 0 && transaction.senderId === transaction.recipientId, positive: transaction.senderId !== $ctrl.account.get().address, negative: transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address }',
+                  data-show-send-modal, data-amount='transaction.amount')
                   lsk.value(amount='transaction.amount')
+                  md-tooltip(md-direction='top', md-delay='350', data-ng-if='transaction.amount > 0') Send {{transaction.amount}} LSK
```
slaweet: This tooltip shows wrong value, e.g. `Send 3000000000 LSK` instead of `Send 30 LSK`

src/components/transactions/transactions.pug
```diff
@@ -32,18 +31,20 @@ md-card.offline-hide
...
+                  md-tooltip(md-direction='top', md-delay='350') Send to this recipient
               td(md-cell)
                 i.material-icons(ng-if='transaction.type === 0 && transaction.senderId === transaction.recipientId') replay
                 i.material-icons.in(ng-if='transaction.senderId !== $ctrl.account.get().address') call_received
                 i.material-icons.out(ng-if='transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address') call_made
               td(md-cell)
-                .amount(ng-class='{ neutral:transaction.type === 0 &&  transaction.senderId === transaction.recipientId, positive: transaction.senderId !== $ctrl.account.get().address, negative: transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address }')
+                .amount(ng-class='{"has-send-modal": transaction.amount > 0, neutral:transaction.type === 0 && transaction.senderId === transaction.recipientId, positive: transaction.senderId !== $ctrl.account.get().address, negative: transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address }',
+                  data-show-send-modal, data-amount='transaction.amount')
```
slaweet: I expected this would set also `, data-recipient-id='transaction.recipientId'`

alihaghighatkhah: What I had in mind was to send the same amount not repeating the transaction. But I'm open to change it if you think that's gonna be more useful.

slaweet: Yes, I think 'repeat transactions' is more useful. But even that is probably still a rare case at the current state of Lisk.

src/components/transactions/transactions.pug
```diff
@@ -32,17 +31,24 @@ md-card.offline-hide
...
+                  md-tooltip(md-direction='top', md-delay='350') Send to this recipient
               td(md-cell)
                 i.material-icons(ng-if='transaction.type === 0 && transaction.senderId === transaction.recipientId') replay
                 i.material-icons.in(ng-if='transaction.senderId !== $ctrl.account.get().address') call_received
                 i.material-icons.out(ng-if='transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address') call_made
               td(md-cell)
-                .amount(ng-class='{ neutral:transaction.type === 0 &&  transaction.senderId === transaction.recipientId, positive: transaction.senderId !== $ctrl.account.get().address, negative: transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address }')
+                .amount(data-ng-if='transaction.type !== 0 || transaction.recipientId !== $ctrl.account.get().address',
+                  data-ng-class='{"has-send-modal": transaction.amount > 0}',
+                  data-show-send-modal, data-amount='transaction.amount', data-recipient-id='transaction.recipientId', class='negative')
```
slaweet: `class='negative'` should be `.negative` after `.amount`
Simillarly in 3 other places in this file

PR #278, opened 2017-05-31
Refactor save passphrase modal into a component - Closes #189

### LINKED ISSUE #189: Refactor new passphrase modal into a component
### Expected behaviour
New passphrase modal should be a separate component so that it can be tested
### Actual behaviour
Currently, new passphrase modal is inside a $mdModal
https://github.com/LiskHQ/lisk-nano/blob/development/src/app/components/login/passphrase.js#L27-L65
and it cannot be easily tested

slaweet:
Closes #189


slaweet:
Apparently, 1.1.0 needs to be updated first

PR #291, opened 2017-05-31
Keep selected network on new account cancellation - Closes #280

### LINKED ISSUE #280: New Account switches network after closing modal
### Expected behaviour
When generating an account and specifying the node or network, it should remain selected when we close the modal.

### Actual behaviour
We can generate the account and close the modal, but it reverts to mainnet

### Steps to reproduce
Open lisk-nano, switch the network, create a new account and close before verifying.

slaweet:


slaweet:
@reyraa I added saving network to cookies, but I'm saving it only after successful login - not to save some invalid address.

Issue #150, opened 2017-04-26
Improve sign and verify message output styling


karmacoma:
#### Sign message

<img width="1134" alt="image" src="https://cloud.githubusercontent.com/assets/8068/25436061/615cf54c-2a92-11e7-8f17-cfa4cdbdd96d.png">

#### Verify message

<img width="1147" alt="image" src="https://cloud.githubusercontent.com/assets/8068/25436068/6cd28df6-2a92-11e7-88b1-c6c3efffa0c7.png">

### Desired appearance

In order to better clarify `input` versus `output` to the user. I propose something like the following which gives a `terminal` feel to them.

```css
textarea {
  color: white;
  background-color: black;
  padding-left: 5px;
  padding-right: 5px;
  border-radius: 5px;
  margin-top: 10px;
  font-family: monospace;
}
```

**Note**: Currently these text areas are editable. We should set them to `readonly`. Alternatively, we can contain them within a `<pre>` element instead.

#### Sign message

<img width="1123" alt="image" src="https://cloud.githubusercontent.com/assets/8068/25436087/7c0f0ede-2a92-11e7-87bd-4e4b258af048.png">

#### Verify message

<img width="1149" alt="image" src="https://cloud.githubusercontent.com/assets/8068/25436092/7fb7f3a2-2a92-11e7-911c-4f257e04213d.png">

### Steps to reproduce

Sign or verify a message using `lisk-nano`.

karmacoma:
I have opened a corresponding issue on `lisk-js` which might help make the actual structure of the returned output clearer: https://github.com/LiskHQ/lisk-js/issues/112

MaxKK:
I re-open this issue as it's something which stood out during the testing. The version Oliver provided looks much, much better. Right now it's a mess. With the CANCEL button it will look very good.

PR #313, opened 2017-06-02
Improve sign/verify message ui

reyraa:
Coses #150


src/components/signVerify/signMessage.js
```diff
@@ -1,4 +1,5 @@
 import lisk from 'lisk-js';
+import './sign-message.less';
```
slaweet: Just for the sake of readability I would do this import also in `src/components/signVerify/verifyMessage.js`
And also rename `sign-message.less` to `signVerifyMessage.less`

reyraa: Thought it's redundant, but I've added.

reyraa:
I don't like the way action bar looks in signMessage, but currently nothing comes into my mind. @slaweet  let me know if you have any suggestions.

slaweet:
@reyraa I tried to improve how the action bar looks, in a commit to this branch.

PR #314, opened 2017-06-02
Check version compatibility on login - Closes #309

### LINKED ISSUE #309: Check version compatibility on login
### Expected behaviour
When Lisk nodes are updated to version 1.0.0 in the future, Lisk Nano should detect that it is trying to connect to an incompatible API version and tell the user to download the new version of Lisk Nano
### Actual behaviour
When Lisk nodes are updated to version 1.0.0 in the future, Lisk Nano will be just broken because of the incompatible API
### Steps to reproduce
Wait for a few months :-) and then try to use Lisk Nano 1.0.0 on Mainnet

slaweet:


slaweet:
To test this you can modify Lisk core node file `api/http/loader.js`
Comment out
```
26     'get /status': 'status',
```
And add:
```
+  30   router.get('/status', function (req, res) {
+  31     return res.status(404).json({success: false});
+  32   });
```

src/services/api/peers.js
```diff
@@ -93,11 +94,24 @@ app.factory('Peers', ($timeout, $cookies, $location, $q, $rootScope) => {
      * @private
      * @memberOf Peer
      * @method check
+     * @notifyUser {bool} - Should an error toast be displayed on error?
      */
-    check() {
+    check(notifyUser) {
```
reyraa: I think it makes perfect sense to notify the user of any error which happens during connections. so this parameter seems to be redundant.

slaweet: The thing is that this should be only on login. After login when this fails, the whole ui get's grey and loading spinner is displayed, so the user is notified in either case. I changed the param name accordingly.

reyraa:
Except the typo, all good. thank you

PR #335, opened 2017-06-06
Add infinity scroll for forged blocks

### LINKED ISSUE #327: Add infinity scroll to "Forged blocks"
### Expected behaviour
That you can just scroll down and new forged blocks appear.

### Actual behaviour
There is a "LOAD MORE" button, even though "Transactions" and "Voting" have infinity scroll.

### Steps to reproduce
Login. Open Forging tab with an active delegate.

alepop:
Resolve #327


slaweet (changes requested):
Works well, apart from when I reach the end of the list - then it keeps trying to load more. 

Note that there is a `count` field in the response to check against if there are more blocks to load.

slaweet (approved):

PR #339, opened 2017-06-07
Conform all numbers with comma - Closes #326

### LINKED ISSUE #326: Conform all numbers with comma
### Expected behaviour
Every bigger number has a separation every 3 digits. (e.g. 1,000 LSK)

### Actual behaviour
Sometimes this is the case, sometimes not.

### Steps to reproduce
- Dashboard: Balance
- Transactions page: Amount numbers
- Forging Page: Block height

yasharAyari:
Use angular number filter in lisk and forging component to fix this issue.


src/components/lsk/lsk.pug
```diff
@@ -1,2 +1,2 @@
-span(ng-bind='$ctrl.amount | lsk')
+span(ng-bind='$ctrl.amount | lsk | number:8')
```
slaweet: `number:8` is also not the answer as we don't want to show the trailing zeros. 
`100.00000000` should be `100`
`100.12300000` should be `100.123`

slaweet: `$ctrl.amount | lsk | number` might work

PR #340, opened 2017-06-07
Use one universal loading bar - Closes #328

### LINKED ISSUE #328: Same loading bar which is all the time at the same location
### Expected behaviour
A standard place for a loading bar which doesn't block any view. I have attached 2 examples which I think would look fine. I prefer the "loading bar at the top".

### Actual behaviour
Currently we have basically have 2 different loading bar styles at 4 different locations.

Transactions: The blue loading bar is about 1cm below the tab itself on the white window.
Voting: The blue loading bar is at the top of the delegate table, but below the column names like "Vote, Rank" etc.
Forging: Right at the top 2mm below the tab itself on the white window.

Peers loading circle in red, overlaid above the network. 

### Steps to reproduce
Open the different tabs or login. Tell me if you need screenshots of the different loading bars.

Loading bar at the top:
new-loading-bar

Loading bar at the bottom:
loadingbar-2

slaweet:


src/components/delegates/delegates.js
```diff
@@ -39,6 +39,7 @@ app.component('delegates', {
       this.votedList = [];
       this.unvoteList = [];
       this.loading = true;
+      this.$scope.$emit('showLoadingBar');
```
reyraa: repeating the same event emit and the manipulating a flag to manage the behaviour is not a good approach. but since it's temporary I'll approve it.

PR #359, opened 2017-06-08
Intro for passphrase generation modals - Closes 331

### LINKED ISSUE #331: After the confirmation of a 1st/2nd passphrase, add another step
### Expected behaviour
We should definitely add an explanation text as a third modal which educates the user that if he loses the passphrase he loses access to his account forever and that once he closes the modal, the password is nowhere saved, unless he took care of it. Maybe add an option to download it locally to the computer.

### Actual behaviour
It just closes and that's it.

### Steps to reproduce
Create a new account or add a second passphrase.

reyraa:
- Adds a step before generating the second passphrase as intro.
 - Moves new account to a modal and creates an intro step before starting the actual generation of passphrase.
Closes #331


slaweet:
Apparently "New account" e2e test has to be updated

PR #384, opened 2017-06-09
Polish voting page - Closes #344

### LINKED ISSUE #344: Polish voting page
- [x] Make the vote checkboxes more prominent (blue like on login page?)
- [x] Highlight voted delegate row in blue (similarly to currently used red/green)
- [x] Change "Confirm vote" to "Confirm" in vote dialog button.

yasharAyari:
- Make the vote checkboxes more prominent (blue like on login page)
- Highlight voted delegate row in blue (similarly to currently used red/green)
- Change "Confirm vote" to "Confirm" in vote dialog button.


src/components/delegates/delegates.less
```diff
@@ -34,7 +34,7 @@ delegates {
     background-color: rgb(255, 228, 220);
   }
   .pending {
-    background-color: #eaeae9;
+    background-color: rgb(204, 231, 236);
```
slaweet: > Highlight voted delegate row in blue (similarly to currently used red/green)

This was supposed to mean something else. Sorry that my description was ambiguous. Pending votes should stay gray, but not pending placed votes should be blue. That is the state when the checkbox is checked and (currently) the background is white. 

Also, I personally find the blue you chose bit too greenish, I like more something like #d6f0ff

src/components/delegates/delegates.less
```diff
@@ -36,6 +36,9 @@ delegates {
   .pending {
     background-color: #eaeae9;
   }
+  .voted{
```
slaweet: `.voted` should be defined before `.pending` in order to have the pending vote gray, not blue.

PR #392, opened 2017-06-12
Fix template issues in forging tab - Closes #391

### LINKED ISSUE #391: Forging tab doesn't show correct template for (non-)delegate accounts
### Expected behaviour
 - For delegates: it should show the their forging information.
 - For non-delegates it should show a massage indicating this tab is for delegates.

### Actual behaviour
It always shows both. and when the account has just registered as delegate the template is broken.

### Steps to reproduce
 - Login.
 - Navigate to `main/forging`.

reyraa:
Closes #391


slaweet:
Unit tests fail

PR #393, opened 2017-06-12
Frameless window

alepop:
Make app window more 💅
- Add frameless window
- Polish loading bar (I think now the best position is at the bottom of the window)

**Before**:
old
**After**:
frameless
loading


slaweet:
Thanks, @alepop, the frameless window looks really cool. 👍 
Though, I noticed that pop-up toast notifications are not going all the way to the screen in this PR:
screen shot 2017-06-13 at 11 25 49
Before it looked like:
screen shot 2017-06-13 at 11 25 14

Also, did you check if there are no other side effects on Windows and Linux?

alepop:
@slaweet ok, i will fix the pop-up. What about loading bar? 
I can not test on Win or Linux 😞

alepop:
@Isabello sounds bad. Maybe set frameless window only for macOS? Need core team decision.

slaweet:
> makes text unselectable on windows and instead moves the whole window when you try to select the text

 This happens also on Mac OS

PR #398, opened 2017-06-13
Open in lisk explorer

alepop:
- Add `openExplorer` directive
- Open transaction on click in Lisk explorer
- Open block on click in Lisk explorer
- Add style for hovered table row
- Open different explorer in main/testnet

### DIFF
--- a/src/components/forging/forging.pug
+++ b/src/components/forging/forging.pug
@@ -67,7 +67,7 @@ md-card.offline-hide
                     th(md-column) Total fee
                     th(md-column) Reward
                 tbody(md-body, infinite-scroll='$ctrl.loadMoreBlocks()', infinite-scroll-distance='1')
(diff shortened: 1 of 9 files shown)


src/components/forging/forging.pug
```diff
@@ -67,7 +67,7 @@ md-card.offline-hide
                     th(md-column) Total fee
                     th(md-column) Reward
                 tbody(md-body, infinite-scroll='$ctrl.loadMoreBlocks()', infinite-scroll-distance='1')
-                  tr(md-row, ng-repeat='block in $ctrl.blocks')
+                  tr(md-row, ng-repeat='block in $ctrl.blocks', data-open-explorer="{{block.id}}", data-url="'/block/'")
```
slaweet: I don't see the point of using two attributes. IMO it would be sufficient and simpler to have just one: `data-open-explorer="/block/{{block.id}}"`

alepop: @slaweet you right.

Isabello:
Branch conflicts, please rebase on your end @alepop

PR #408, opened 2017-06-14
Load 20 transaction instead 10 - Closes #402

### LINKED ISSUE #402: Remove the "Load More" link at the bottom of the first few transactions
I think it would be more convenient if the additional transactions are loaded automatically when scrolled down, as on the Votings and Forging pages. Everything should work the same way

alepop:
close #402


slaweet:
Thanks @alepop for the PR. Also unit tests need to be updated.

reyraa (approved):

slaweet (approved):

PR #411, opened 2017-06-14
Unify "not enough funds" errors - Closes #405

### LINKED ISSUE #405: Unify "not enough funds" errors
### Expected behaviour
All transaction types should check balance at the beginning and display a unified error message.
### Actual behaviour
- Send: "Insufficient funds" error message under the "amount" input
- Vote: red error toast saying " Account does not have enough LSK: 9192292151089507854L balance: 0" after clicking "Confirm" button
- Register as delegate: "Account does not have enough LSK: 9192292151089507854L balance: 0" message after clicking "Register" button
- Register second passphrase: "You have insufficient funds to register a second passphrase." Error alert dialog after clicking "Register" button

slaweet:


src/components/fee/fee.js
```diff
@@ -0,0 +1,29 @@
...
+  template: '{{$ctrl.text}}',
+  bindings: {
+    fee: '<',
+  },
+  controller: class fee {
+    constructor($scope, Account, lsk, $element) {
+      this.account = Account;
+      const notEnoughtLSK = lsk.normalize(this.account.get().balance) < this.fee;
+
+      this.text = notEnoughtLSK ?
```
reyraa: Please fix the typo. Also fix it in L23.

reyraa: Also I would suggest `insufficientFunds`.

src/filters/notEnoughBalance.js
```diff
@@ -0,0 +1,8 @@
+
+/**
+ * This filter
+ *
+ * @module app
+ * @submodule notEnoughBalance
+ */
+app.filter('notEnoughBalance', (Account, lsk) => amount => lsk.normalize(Account.get().balance) < amount);
```
reyraa: And this name can be `FundsSufficiency` as in _Filter by funds sufficiency_.

PR #418, opened 2017-06-15
Use <a> element to parse url instead of Regexp

### LINKED ISSUE #417: Using wrong URL as custom node results in endless loop
Lisk Nano v1.0.0-rc.4

Lisk Nano gets into an endless loop, if it's trying to connect to an invalid / not reachable node.

### Expected behaviour
"Couldn't connect to node."-message and the ability to change URL and try login again.

### Actual behaviour
Nano loops, trying to connect to wrong URL. Needs restart.

### Steps to reproduce
Instead of https://liskwallet.punkrock.me, I just typed in liskwallet.punkrock.me - Nano is trying to connect on port 80 (I assume) endless times. I still can change the URL now, but I'm not able to click on Login again. I would have to close Nano and try again.

alepop:
There an error when parsing the wrong URL with regexp.
fix #417

### DIFF
--- a/src/services/api/peers.js
+++ b/src/services/api/peers.js
@@ -1,5 +1,5 @@
 import lisk from 'lisk-js';
-
+import urlParser from '../../util/urlParser';
 /**
  * This factory provides methods for communicating with peers. It exposes
  * sendRequestPromise method for requesting to available endpoint to the active peer,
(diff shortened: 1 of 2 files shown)


slaweet (changes requested):
Thank you @alepop for the PR. It's much better than my original solution, though it still doesn't work well on the reported input (`liskwallet.punkrock.me`). `<a>` element treats it as local path. In electron wrapper, it connects to a mainnet peer. On `localhost:8080` it tries to connect to `localhost:8080`.

alepop:
@slaweet yep. It is a basic behavior for nonvalid URL format. if you pass a string that not represent the URL the `a` element use current URL from the browser.

PR #428, opened 2017-06-20
Show balance in another currency - Closes #362

### LINKED ISSUE #362: Show balance in another currency too
## Feature Request

whitemagnum on lisk chat asked about seeing your lisk balance in another currency like USD/EURO/etc.  I think it would be a nice option to add like on the explorer where there is a dropdown and you can pick a currency to see the converted values.

alepop:
- Add PriceApi service that fetching tickers from lisk-explorer API.
- Ticker data will update on sync event
- Add dropdown to switch between currencies.
- Add `currency` filter and update `lsk` component.
- Use `Intl.NumberFormat` for formatting numbers. 

Closes #362


otint:
@karmacoma @alepop @slaweet it looks like this feature was coded and ready to go -- what happens next?

alepop:
@otint it was implemented for the old code base. now after migration was complete, I think that I can create another PR. What do you think @karmacoma, @slaweet ?

slaweet:
@alepop sorry for late response. You are welcome to create another PR for this feature.
 
As for the details of the feature, I think it's better to show the fiat value in addition to the LSK value, not instead of LSK value. 

Also, I wouldn't bother with displaying fiat values on the list of transactions, because it should probably use the conversion ratio from the time the transaction was made and that is too complicated.

PR #433, opened 2017-06-22
Account logic - Closes #429

### LINKED ISSUE #429: Implement account logic in React
We need to have the logic of fetching and maintaining account information implemented. This includes all actions, reducers and every required utilities along with their unit tests.

reyraa:
- Adds actions to set and reset account details along with required utilities.
 - Add reducers to manage the state for account info.

### DIFF
--- a/src/components/counter.js
+++ b/src/components/counter.js
@@ -7,6 +7,8 @@ import { createStore } from 'redux';
 import { Provider, connect } from 'react-redux';
 import PropTypes from 'prop-types';
 
+document.addEventListener('syncTick', e => console.log(e));
+
 /**
 * @description this is a reducer for handling counter state
 * @param {number} state - current state of our component
@@ -47,7 +49,7 @@ Counter.propTypes = {
   reset: PropTypes.func,
 };
 
-const ReduxConter = () => (
+const ReduxCounter = () => (
   <Provider store={createStore(counterReducer)}>
     <ConnectedCounter />
   </Provider>
@@ -80,4 +82,4 @@ const mapStateToProps = state => ({ value: state });
 
 const ConnectedCounter = connect(mapStateToProps, mapDispatchToProps)(Counter);
 
-export default ReduxConter;
+export default ReduxCounter;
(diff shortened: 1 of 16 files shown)


src/components/counter.js
```diff
@@ -7,6 +7,8 @@ import { createStore } from 'redux';
 import { Provider, connect } from 'react-redux';
 import PropTypes from 'prop-types';
 
+document.addEventListener('syncTick', e => console.log(e));
+
```
alepop: Redux logger will show you all actions that will be triggered. Is this listener needed?

reyraa: no, that was just for dev purpose which I forgot to delete

slaweet:
Overall great work.
So far I noticed two issues:
- Jenkins passes even though there are 2 eslint errors in Jenkins
- `metronome.js` has low test coverage.

src/utils/polyfills.test.js
```diff
@@ -0,0 +1,40 @@
+import { expect } from 'chai';
+import { deepEquals } from './polyfills';
+
+describe('Polyfills', () => {
```
slaweet: Here I'm missing one level of describe:
`describe('deepEquals', () => {`

src/main.js
```diff
@@ -2,9 +2,23 @@
 import React from 'react';
 import ReactDom from 'react-dom';
 import { BrowserRouter as Router, Route, Link } from 'react-router-dom';
-import ReduxConter from './components/counter';
+import ReduxCounter from './components/counter';
+import Metronome from './utils/metronome';
 
-const App = () => (
+class App extends React.Component {
```
slaweet: this file (`src/main.js`) is not included in the test coverage report. Apparently, the report covers only files inside at least second level folders. This file has no tests

reyraa: This tic ket is not related to this file

Isabello:
Can fix this. I think is a bug in the way it is executing eslint

src/utils/polyfills.test.js
reyraa: Done

Issue #435, opened 2017-06-23
Add a material design based framework to react boilerplate


yasharAyari:
This framework should have these components :

1. dialog box
2. tab
3. selectable table
4. tooltip
5. grid base layout system

slaweet:
Done in https://github.com/LiskHQ/lisk-nano/pull/436

PR #436, opened 2017-06-23
Add material design frame work to react boilerplate - cloase #435 

yasharAyari:


src/index.html
```diff
@@ -2,7 +2,9 @@
 <html>
     <head>
         <meta charset="utf-8">
-        <title>Webpack fun</title>
+        <title>isk nano</title>
+        <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+        <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet">
```
slaweet: The fonts should be stored locally.

package.json
```diff
@@ -29,11 +29,14 @@
     "bitcore-mnemonic": "=1.1.1",
     "lisk-js": "=0.4.2",
     "moment": "=2.15.1",
+    "postcss": "^6.0.2",
```
slaweet: please save all dependencies with `=` instead of `^`, e.g. "postcss": "=6.0.2",`

PR #463, opened 2017-07-10
Remove boilerplate example components - Closes #462

### LINKED ISSUE #462: Remove boilerplate example components
We've added `src/component/static` and `src/component/counter` components as examples while creating the boilerplate. Now that there are actual components ready we can delete them.

reyraa:
Closes #462


src/main.js
```diff
@@ -2,51 +2,32 @@
...
-            <Route path="/static" component={Static} />
-            <Route path="/counter" component={this.ReduxCounter} />
-          </div>
-        </Router>
-      </section>
-    );
-  }
-}
+  return (
+    <section className={styles['body-wrapper']}>
```
alepop: Maybe `styles.bodyWrapper` is better?

reyraa: Though I've removed this file, I'd also suggest using camelCase.

PR #481, opened 2017-07-12
Fix notification about negative balance - Closes #477

### LINKED ISSUE #477: Lisk Nano notified me of a negative balance
### Expected behaviour
Only notifies me of incoming transaction.

### Actual behaviour
When I forged a block, I received a notification stating ``You've just received -55.xxx LSK``. The notification included the 5 Forged LSK + the 50ish I was sending to someone else.

### Steps to reproduce
Forge a block and send transactions at the same time

alepop:
Closes #477

### DIFF
--- a/src/components/main/main.js
+++ b/src/components/main/main.js
@@ -109,8 +109,10 @@ app.component('main', {
             delete res.publicKey;
           }
 
-          if (res.balance > this.account.get().balance) {
-            const amount = res.balance - this.account.get().balance;
+          const currentBalance = this.account.get().balance;
+
+          if (res.balance > currentBalance) {
+            const amount = res.balance - currentBalance;
             this.notify.about('deposit', amount);
           }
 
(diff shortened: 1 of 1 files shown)


slaweet:
@alepop can you please add some description of how/why the bug happens? I don't see what difference do your changes make, given that JavaScript is single-threaded.

alepop:
@slaweet I know that the JavaScript in single-threaded and so this bug confused me. I can not find an answer to this error except as a possible balance change after the check, but how I not understand. And I could not reproduce this behavior. I think that the variable cache may help.

slaweet:
Even more explicit that negative balance cannot happen would be to do the `const amount = ...` outside the if and then do `if(amount>0)`

slaweet:
I still don't see how this issue could have happened, but let's merge this in and see if the issue happens again.

PR #490, opened 2017-07-16
Migrate transactions tab to react - Closes #349

### LINKED ISSUE #349: Migrate transactions tab to React

yasharAyari:


src/components/transactions/index.js
```diff
@@ -0,0 +1,62 @@
...
+        });
+      })
+      .catch(error => console.error(error.message));
+    }
+  }
+
+  render() {
+    return (
+      <div className='box'>
+        <table className={tableStyle.table}>
```
slaweet: Using the react toolbox `<Table>` component should be quite straightforward and will allow us to get rid of `className={tableStyle.table}` in may places to make the code cleaner. Here's the docs: https://github.com/react-toolbox/react-toolbox/tree/dev/components/table

src/components/transactions/index.js
```diff
@@ -0,0 +1,62 @@
...
+    this.loadMore();
+  }
+
+  loadMore() {
+    if (this.state.loadMore && this.state.length > this.state.offset) {
+      this.setState({ loadMore: false });
+      transactions(this.props.activePeer, this.props.address, 20, this.state.offset)
+      .then((res) => {
+        const list = res.transactions.map(transaction => (
+          <TransactionRow address={this.props.address}
```
slaweet: I think that the state should contain the list of transactions returned from the server, not list of React components.

src/components/transactions/index.js
```diff
@@ -0,0 +1,62 @@
...
+            key={transaction.id}
+            tableStyle={tableStyle}
+            value={transaction}>
+          </TransactionRow>
+        ));
+        this.setState({
+          transactions: this.state.transactions.concat(list),
```
slaweet: The transactions list should be in redux, so that we can add there "pending" transactions from other components - send, vote, etc.

src/components/transactions/index.js
yasharAyari: I think in this component we don't need any parts of react-toolbox table features such as sorting or selectable row, furthermore, we will remove react-toolbox and other material design related framework in near future after we get new design.So, it is better not to use react-toolbox  components if we don't need them.

src/components/transactions/index.js
yasharAyari: I think list of pending transactions should store in redux store and they come from voting component which hasn't been developed yet.So we can't implement them now.We can't store list of transactions in redux store because they come from server.

src/components/transactions/transactionType.test.js
```diff
@@ -0,0 +1,90 @@
...
+  it('expect text of TransactionType to be equal "props.sederId"', () => {
+    const inputValue = {
+      type: 0,
+      senderId: '1085993630748340485L',
+    };
+    const expectedValue = `<div class="">${inputValue.senderId}</div>`;
+    const wrapper = shallow(<TransactionType {...inputValue} />);
+    expect(wrapper.find(TooltipWrapper).html()).to.be.equal(expectedValue);
+  });
+//   it('expect amount to have className "transactions__inButton"', () => {
```
reyraa: Please use `it.skip` instead of commenting

src/components/transactions/index.js
reyraa: I think we just need to keep track of our pending transactions in store. keeping the confirmed transactions list is redundant. and about the pending transactions, I'd suggest implementing it in a separated ticket after Send or Vote are done.

slaweet: I don't see why 
> We can't store list of transactions in redux store because they come from server.
Why cannot we dispatch `updateTransactions` action with the data returned in `then()`? Like Ali does it here: https://github.com/LiskHQ/lisk-nano/pull/493/files#diff-d2128c41f8409575d6663f2279016a25R31

reyraa: Keeping transactions in store doesn't add any useful data and just enlarges our state. I can't find a good reason for storing it.

src/components/transactions/index.js
reyraa: It doesn't matter if we want to delete or not, using `Table` is a neat approach comparing to class name. in parallel with semantic HTML.

src/components/transactions/index.js
slaweet: I think that the reason to dispatch `updateTransactions` action is that there will be also a reducer that removes them from pending transactions. I agree with handling 'pending transactions' in a separate issue

Also when I switch from and back to "transactions" tab it can immediately show the list from redux and then check for any updates.

src/components/formattedNumber/index.js
```diff
@@ -1,20 +1,28 @@
...
 /**
  *
  * @param {*} num - it is a number that we want to format it as formatted number
  * @return {string} - formatted version of input number
  */
 const formatNumber = (num) => {
-  num = parseFloat(num);
-  const sign = num < 0 ? '-' : '';
-  const absVal = String(parseInt(num = Math.abs(Number(num) || 0), 10));
+  let normalizeNum = parseFloat(normalize(num));
```
slaweet: `normalize(num)` IMO should not be in `<FormattedNumber />` component
I'm using `<FormattedNumber />` to format "Block height" in "Forged blocks", which is a number, but not LSK value. 
I think we need another component `<LiskAmount val={val} />` that will render something like `LiskAmount =
 props => (<FormattedNumber val({normalize(props.val)}/> LSK)`. 
But I will create a separate issue for it. This is already big enough.

PR #493, opened 2017-07-17
Update account regularly on beat events - Closes #488

### LINKED ISSUE #488: Update account regularly on beat events
### Expected behaviour
We have to make regular Api calls to keep account update. I suggest implementing this as a part of account component.

### Actual behaviour
Currently we're reading from a mock data.

reyraa:
Binds event listener to `beat` events coming from `metronome` to update `state.account` using `Account` Api utility.
Closes #488


src/components/app/index.js
```diff
@@ -11,6 +11,8 @@ import Metronome from '../../utils/metronome';
 import { setActivePeer } from '../../utils/api/peers';
 import { accountUpdated } from '../../actions/account';
 import Dialog from '../dialog';
+// temporary
```
slaweet: It might be better to write these comments more descriptive, something like:

`// TODO: next line should be moved to XXX once we setup YYY.`

But maybe it's not worth in this case, depends on how soon do you expect it to be resolved.

src/store/reducers/peers.js
```diff
@@ -1,16 +1,21 @@
...
+ * @param {Array} state - the current state object
+ * @param {Object} action - The action containing type and data
+ *
+ * @returns {Object} - Next state object
  */
 const peers = (state = {}, action) => {
   switch (action.type) {
     case actionTypes.activePeerSet:
-      return Object.assign({}, state, { active: action.data });
+      return Object.assign({}, state, { data: action.data });
```
reyraa: In this case, data makes perfect sense. There might be tens of peers available. but we just need to keep the tracks of the active one. so, I'm storing the data I get from lisk-js as its `data` and the potential configs we might add in lifecycle of Nano as `state`.

reyraa: But still, if you think you're not happy with it, let know to change it

PR #498, opened 2017-07-19
Migrate Login page - Closes #347

### LINKED ISSUE #347: Migrate login page to React
- [x] Login container component
- [x] Login form with all functionality

reyraa:
- Creates the login page
   - Form validation
   - Sets account and active peer on submission
   - Doesn't include create account dialog
 - Removes `setActivePeer` and `resetActivePeer` from utilities and adds the logic to actions.
Closes #347


src/components/login/index.js
```diff
@@ -1,21 +1,25 @@
...
-      <h1>{this.title}</h1>
-    );
-  }
-}
-export default Login;
+const LoginComponent = () => (
+  <section>
+    <div className={`box ${styles.paddingTop}`}>
+      <div className={`${grid.row}`}>
+        <div className={grid['col-xs-2']}></div>
```
yasharAyari: Please remove two extra '<div className={grid['col-xs-2']}></div>' and add "grid[‘center-xs’]" class to your row. Because we use a flexbox based grid system we don't need those extra div any more.You can find more about it here : http://flexboxgrid.com/

PR #502, opened 2017-07-21
Create <LiskAmount val={val} /> React component - Closes #501

### LINKED ISSUE #501: Create <LiskAmount val={val} /> React component
`normalize(num)` should not be in `<FormattedNumber />` component
Because of it `<FormattedNumber />` cannot be used to format "Block height" in "Forged blocks", which is a number, but not LSK value. 
We need component `<LiskAmount val={val} />` that will render something like `LiskAmount =
 props => (<FormattedNumber val({normalize(props.val)}/> LSK)`.

slaweet:
Closes #501

### DIFF
--- a/src/components/liskAmount/index.test.js
+++ b/src/components/liskAmount/index.test.js
@@ -3,9 +3,9 @@ import { expect } from 'chai';
 import { mount } from 'enzyme';
 import LiskAmount from './';
 
-describe('<LiskAmount />', () => {
+describe('LiskAmount', () => {
   const normalizeNumber = 100000000;
-  it('should render "12932689.645" as "12,932,689.645"', () => {
+  it('should normalize "12932689.645" as "12,932,689.645"', () => {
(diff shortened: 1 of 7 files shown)


src/components/liskAmount/index.test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { expect } from 'chai';
+import { mount } from 'enzyme';
+import LiskAmount from './';
+
+describe('<LiskAmount />', () => {
```
reyraa: Please remove HTML annotations

src/components/liskAmount/index.test.js
```diff
@@ -0,0 +1,14 @@
...
+  const normalizeNumber = 100000000;
+  it('should render "12932689.645" as "12,932,689.645"', () => {
```
reyraa: Please use a more descriptive title. I'd suggest:
should normalize "12932689.645" as "12,932,689.645"

PR #509, opened 2017-07-25
Enhancements in routing - Closes #499

### LINKED ISSUE #499: Enhancements in routing
### Expected behaviour
 - After logging in, clicking back button should not #navigate to login page.
 - Reloading should sign out the user and redirect to login page
 - Fix unit tests.
### Actual behaviour
 - Currently clicking on back button navigates to login page, even if already logged in.
 - reloading breaks the app ( if using # in route structure helps fixing this, it's fine)
 - We've skipped existing route tests since navigating directly to those routes is not possible. we need to navigate to Login, and call onLoginSubmission with proper inputs, or find an approach to store active peer and account info on the store.

### Steps to reproduce
Run e2e tests for the skipped test in `src/components/app.js`

alepop:
* Added `<PrivateRoute />` component.
* Added `babel` stage-3
* Fixed `Metronome.init()`. Moved it from `App` component. 
* Rewrite `App` component tests. They are not cover all cases because it is hard to test root component with many dependencies (Provider, Router)
close #499


webpack.config.js
```diff
@@ -70,7 +70,7 @@ module.exports = (env) => {
           exclude: /node_modules/,
           loader: 'babel-loader',
           options: {
-            presets: ['es2015', 'react'],
+            presets: ['es2015', 'react', 'stage-3'],
```
slaweet: What do you use `stage-3` for? I tried to read some basics about what it is, but I cannot find what new features you use.

slaweet: Now I see it is because of `{ component: Component, ...rest }`

src/components/app/index.js
```diff
@@ -11,32 +11,21 @@ import Metronome from '../../utils/metronome';
...
+const metronome = new Metronome();
+metronome.init();
 
-  return (
-    <section className={styles['body-wrapper']}>
-      <Header />
-      <main className=''>
-        <Route path="/main" render={({ match }) => (
-          <main className=''>
-            <Account />
```
yasharAyari: We don't need to render Account component and Link components every time that we change the route.They are the same in forging and voting and transaction and rendering them over and over is useless and it will decrease performance of our application .So, we must use nesting routes to solve this problem.why do you remove nested routes?

src/components/defaultLayout/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import Account from '../account';
+import PrivateRoute from '../privateRoute';
+
+const DefaultLayout = ({ component: Component, ...rest }) => (
+  <PrivateRoute {...rest} component={ matchProps => (
+    <main>
+      <Account />
```
yasharAyari: We don't need to render Account component and Link components every time that we change the route.They are the same in forging and voting and transaction and rendering them over and over is useless and it will decrease performance of our application .So, we must use nesting routes to solve this problem.why do you remove nested routes?

alepop: The Links component was moved on @slaweet request. I think that the performance change here is negligible especially if you will move the logic from account component and keep it dumb and allow only render the passed props.  I do not know how the app will be looked after rebranding, and I think that the layout can help combine some components for differents views.
If you think that is a bad solution than I can return the nested route.

yasharAyari: I noticed you render Account component and Link components every time changing the route,  I think rending them once will help improving performance. to do so, I’d suggest find a way to keep those components mounted once or adapting unit tests. nested routes are used in many apps. there should be some other applications facing this situation.

alepop: @yasharAyari fixed

src/components/privateRoute/index.test.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { expect } from 'chai';
+import { mount } from 'enzyme';
+import { MemoryRouter, Route } from 'react-router';
+import { PrivateRouteRender } from './index';
+
+const Public = () => <h1>Public</h1>;
+const Private = () => <h1>Private</h1>;
+
+describe('<PrivateRouteRender />', () => {
```
slaweet: Please use just `PrivateRouteRender` to make it consistent with other Components' `describe`s

src/components/defaultLayout/index.js
yasharAyari: good job. thank you.

PR #519, opened 2017-07-26
Setup React toaster - Closes #486

### LINKED ISSUE #486: Setup React toaster library
### Expected behaviour
We use toast notifications in several places, so there should be a library to handle those.
This toaster lib looks good so far: https://github.com/diegoddox/react-redux-toastr though it's using sass for styles and I'm not sure how to use those with postcss.
### Actual behaviour
There is no toaster in the current state od react migration. Even React toolbox doesn't have a toaster

slaweet:
Closes #486
But not using `react-redux-toaser`. 
Instead, using http://react-toolbox.com/#/components/snackbar because we are already using react toolbox so there's no point in adding another library.


src/actions/toaster.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+export const errorToastDisplayed = ({ type = 'error', ...rest }) =>
+  toastDisplayed({ type, ...rest });
+
+/**
+ * An action to dispatch to hide a toast
+ *
+ */
+export const toastHidden = () => ({
```
reyraa: This doesn't handle multiple toast. we're using toasts to notify user about many changes, like sign message and network status changes. There's a good chance we'll need to show multiple - 2 is more realistic - toasts at the same time.
Please change this to include `id` and tweak other parts of code accordingly.
thanks

slaweet: Toasts are by material design specs limited to one.

> "They should not be persistent or be stacked, as they are above other elements on screen."
> http://www.google.com/design/spec/components/snackbars-toasts.html#snackbars-toasts-usage

In Angular Material, it wasn't possible to stack them either.

Issue #505, opened 2017-07-21
Setup React toolbox tabs for Transactions/Voting/Forging


slaweet:
### Expected behaviour
Tabs use http://react-toolbox.com/#/components/tabs with the style adjustments we had in Angular
### Actual behaviour
There are just plain links

slaweet:
Also "Forging" tab should only be visible if `account.isDelegate`

PR #523, opened 2017-07-27
Re-enable e2e tests for features already working in React - Closes #507

### LINKED ISSUE #507: Re-enable e2e tests for features already working in React
### Expected behaviour
There is more functionality already implemented in React but with disabled e2e tests - Login, Transactions, Account and more. 
### Actual behaviour
Only sign/verify message has activated e2e tests.

slaweet:
Closes #507


src/components/app/index.js
```diff
@@ -23,9 +23,11 @@ const App = () => (
       <PrivateRoutes path='/main' render={ ({ match }) => (
         <main>
           <Account />
-          <Link to={`${match.url}/transactions`}>Transactions</Link>
-          <Link to={`${match.url}/voting`}>Voting</Link>
-          <Link to={`${match.url}/forging`}>Forging</Link>
+          <section className='main-tabs'>
```
yasharAyari: I think it is better to use nav tag instead of section here.

features/menu.feature
```diff
@@ -82,29 +81,25 @@ Feature: Top right menu
          -----END LISK SIGNED MESSAGE-----
          """
 
-  @ignore
   Scenario: should allow to exit sign message dialog
```
reyraa: Please reflect the difference of this test with the next one in their titles

src/components/app/index.js
slaweet: Good point. But not worth fixing because it will be replaced with `<Tabs>` in https://github.com/LiskHQ/lisk-nano/issues/505 anyway.

features/menu.feature
```diff
@@ -82,29 +81,25 @@ Feature: Top right menu
...
     Given I'm logged in as "any account"
     When I click "sign message" in main menu
     And I click "cancel button"
```
reyraa: When I remove this line, I expect the test to fail, but it passes. apparently this line doesn't affect the test result.
Same goes for the next 2 scenarios testing x and cancel buttons close our sign and verify message dialogs.

PR #526, opened 2017-07-27
Implement custom alert dialogs in React - Closes #515

### LINKED ISSUE #515: Implement custom alert dialogs in React
### Expected behaviour
Displaying an alert dialog should be a simple function call `successAlert({title = 'Success', text})` or `errorAlert({title = 'Error', text})` which puts some data into redux and based on that the modal dialog from react toolbox is displayed.

Also: use it in Send dialog.
### Actual behaviour
There is no custom alert dialogs

slaweet:
Closes #515

### DIFF
--- a/src/components/dialog/alert.js
+++ b/src/components/dialog/alert.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import Button from 'react-toolbox/lib/button';
+import grid from '../../../node_modules/flexboxgrid/dist/flexboxgrid.css';
+
+
+const Alert = props => (
+  <div>
+    <p>{props.text}</p>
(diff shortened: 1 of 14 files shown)


src/components/dialog/alert.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import Button from 'react-toolbox/lib/button';
+import grid from '../../../node_modules/flexboxgrid/dist/flexboxgrid.css';
```
reyraa: For importing from npm modules, there's not need to use relative path. `flexboxgrid/dist/flexboxgrid.css';` would work fine.

PR #534, opened 2017-07-31
Implement click to send functionality in React - Closes #516

### LINKED ISSUE #516: Implement click to send functionality in React
### Expected behaviour
There should be several places where click opens Send dialog
- Account balance: sets maximum amount
- Transaction row "From/To": sets the same recipient
- Transaction row "Amount" of outgoing transaction: sets the same recipient and amount

### Actual behaviour
None of the above is there, yet.

slaweet:
Closes #516

### DIFF
--- a/src/components/send/clickToSendComponent.js
+++ b/src/components/send/clickToSendComponent.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import styles from './send.css';
+import Send from '../send';
+import { fromRawLsk } from '../../utils/lsk';
+
+
+const ClickToSendComponent = props => (
+  props.disabled ?
+    props.children :
+    <span className={styles.clickable}
+      onClick={() => (props.setActiveDialog({
+        title: 'Send',
+        childComponent: Send,
+        childComponentProps: {
+          amount: props.rawAmount ? fromRawLsk(props.rawAmount) : props.amount,
+          recipient: props.recipient,
+        },
+      }))}>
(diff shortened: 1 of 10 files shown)


src/components/send/clickToSendComponent.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import styles from './send.css';
+import Send from '../send';
+import { fromRawLsk } from '../../utils/lsk';
+
+
+const ClickToSendComponent = props => (
```
reyraa: If you please rename the Send component to SendForm, then you can use Send for this component. TBH, I don't like the name

src/components/send/clickToSendComponent.js
```diff
@@ -0,0 +1,23 @@
...
+  props.disabled ?
+    props.children :
+    <span className={styles.clickable}
+      onClick={() => (props.setActiveDialog({
+        title: 'Send',
+        childComponent: Send,
+        childComponentProps: {
+          amount: props.rawAmount ? fromRawLsk(props.rawAmount) : props.amount,
```
reyraa: This coms handy. Bingo

src/components/send/clickToSendComponent.js
slaweet: I gave it some thought and I couldn't think of another name that exactly expresses what the component does. I like the name

PR #538, opened 2017-07-31
Migrate global loading bar to React - Closes #494

### LINKED ISSUE #494: Implement global loading bar React component
### Expected behaviour
We need the global loading bar component and some actions and util to toggle the state.
I think that the state should be a list/dict of things that are in progress and the actions to add and remove them.
### Actual behaviour
We don't have it yet in react

slaweet:
Closes #494

### DIFF
--- a/src/utils/api/peers.js
+++ b/src/utils/api/peers.js
@@ -1,12 +1,16 @@
+import { loadingStarted, loadingFinished } from '../../utils/loading';
+
 /* eslint-disable  */
 export const requestToActivePeer = (activePeer, path, urlParams) =>
   new Promise((resolve, reject) => {
+    loadingStarted(path);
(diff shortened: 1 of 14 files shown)


src/utils/api/peers.js
```diff
@@ -1,12 +1,16 @@
+import { loadingStarted, loadingFinished } from '../../utils/loading';
+
 /* eslint-disable  */
 export const requestToActivePeer = (activePeer, path, urlParams) =>
   new Promise((resolve, reject) => {
+    loadingStarted(path);
```
yasharAyari: I think we should use loading actions inside our component instead of using them inside our api utils

yasharAyari: I think we should use loading actions inside our component instead of using them inside our api utils

PR #542, opened 2017-08-01
Fix header component - Closes #506

### LINKED ISSUE #506: Fix Header component
### Expected behaviour
- Only logo is displayed if user not logged in.
- Logout button works.
- Bigger margin-bottom and smaller margin-right (align open drop-down menu)

alepop:
- Add `<PrivateWrapper />` component
- Handle logout action on Logout button click
- Align drop-down menu
close #506

### DIFF
--- a/src/components/header/header.css
+++ b/src/components/header/header.css
@@ -16,3 +16,7 @@
 .material-icons{
     font-size: 24px !important;
 }
+
+.menu {
+    right: -16px !important;
+}
--- a/src/components/privateWrapper/index.js
+++ b/src/components/privateWrapper/index.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+export const PrivateWrapperComponent = ({ isAuthenticated, children }) => (
+    isAuthenticated && <span>{ children }</ span>
+);
+
+const mapStateToProps = state => ({
+  isAuthenticated: !!state.account.publicKey,
+});
+
+export default connect(mapStateToProps)(PrivateWrapperComponent);
--- a/src/components/transactions/transactionsComponent.js
+++ b/src/components/transactions/transactionsComponent.js
@@ -39,7 +39,7 @@ class Transactions extends React.Component {
           length: parseInt(res.count, 10),
         });
       })
-      .catch(error => console.error(error.message));
+      .catch(error => console.error(error.message)); //eslint-disable-line
     }
   }
 
--- a/src/components/account/accountComponent.js
+++ b/src/components/account/accountComponent.js
@@ -11,9 +11,17 @@ import { getAccountStatus } from '../../utils/api/account';
  * @param {object} props - include properties of component
  */
 class AccountComponent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.update = this.update.bind(this);
+  }
   componentDidMount() {
     this.update();
-    document.addEventListener('beat', this.update.bind(this));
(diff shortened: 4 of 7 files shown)


src/components/privateWrapper/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+export const PrivateWrapperComponent = ({ isAuthenticated, children }) => (
+    isAuthenticated && <span>{ children }</ span>
+);
+
+const mapStateToProps = state => ({
+  isAuthenticated: !!state.account.publicKey,
```
slaweet: this should be something else then `publicKey` because now it shows the haeder right when I click the "Login" button, but it should only after the server request is finished and it is redirected from login page.

src/components/transactions/transactionsComponent.js
```diff
@@ -39,7 +39,7 @@ class Transactions extends React.Component {
           length: parseInt(res.count, 10),
         });
       })
-      .catch(error => console.error(error.message));
+      .catch(error => console.error(error.message)); //eslint-disable-line
```
slaweet: I see that finally somebody got annoyed enough by the warning in console :-) Though I would prefer to disable the specific rule. In this case a believe it is `//eslint-disable-line no-console`

src/components/header/header.css
```diff
@@ -16,3 +16,7 @@
 .material-icons{
     font-size: 24px !important;
 }
+
+.menu {
+    right: -16px !important;
```
slaweet: This could be less (maybe `right: -8px !important;`) and the put the rest on `.wrapper`.
`.wrapper` has too much space on the right compared to nano 1.0.0 and not enough at the bottom. 

I mean something like this is missing
```
.wrapper {
  margin-right: -8px;
  margin-bottom: 16px;
}
```

src/components/privateWrapper/index.js
alepop: currently, all private routes look to the same params.

src/components/header/header.css
alepop: `menu` is a theme param, and it passes to the component and merges with other styles. it is so hard to style `Menu` component through `IconMenu` component.

src/components/privateWrapper/index.js
slaweet: In private routes, it makes no user-visible difference. But on the login page if user is on slower network, it can longer time for the request and only then user is redirected from login page. These login buttons should not be on login page.

src/components/header/header.css
slaweet: I mean `.wrapper` which is defined and used in our `header` component. Nothing extra in `IconMenu`
https://github.com/alepop/lisk-nano/blob/2c484cfd1565634b4f682fbf207ee2bc3ce3bb42/src/components/header/header.css#L1-L4
https://github.com/alepop/lisk-nano/blob/2c484cfd1565634b4f682fbf207ee2bc3ce3bb42/src/components/header/headerElement.js#L12-L13

src/components/privateWrapper/index.js
slaweet: This was fixed in the meantime by some other changes.

src/components/header/header.css
slaweet: This is still an issue

alepop: Do i need to fix it in other pr?

slaweet: Yes, please.

PR #543, opened 2017-08-01
Register as delegate - Closes #354

### LINKED ISSUE #354: Migrate Register as delegate to React

reyraa:
Closes #354


src/components/header/headerElement.js
```diff
@@ -20,7 +21,13 @@ const HeaderElement = props => (
         theme={styles}
       >
         <MenuItem caption="Register second passphrase" />
-        <MenuItem caption="Register as delegate" />
+        <MenuItem caption="Register as delegate"
+          className={(props.account.isDelegate) ? styles.hidden : ''}
```
slaweet: I see the PR is in progress, but I just noticed this one thing. IMO more elegant way to conditionally hide an element in react is the ternary operator - `{shouldShow ? <Element /> : null }`.

reyraa: On a small component with no reprinting effect, there's no difference. I'll change anyways.

src/components/header/header.css
```diff
@@ -20,3 +20,6 @@
 .menu {
     right: -16px !important;
 }
+.hidden {
```
slaweet: `.hidden` is not used.

src/components/header/headerElement.test.js
```diff
@@ -18,9 +18,11 @@ describe('HeaderElement', () => {
   beforeEach(() => {
     const mockInputProps = {
       setActiveDialog: () => { },
+      account: {},
     };
     propsMock = sinon.mock(mockInputProps);
-    wrapper = shallow(<HeaderElement setActiveDialog={mockInputProps.setActiveDialog} />);
+    wrapper = shallow(<HeaderElement account={mockInputProps.account}
+      setActiveDialog={mockInputProps.setActiveDialog} />);
```
slaweet: can be simplified to `wrapper = shallow(<HeaderElement { ...mockInputProps }/>);`

src/components/registerDelegate/index.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+  const account = {
+    isDelegate: false,
+    address: '16313739661670634666L',
+    username: 'lisk-nano',
+  };
+
+  const store = {
+    dispatch: () => {},
+    subscribe: () => {},
```
slaweet: You can use `redux-mock-store` like here:
https://github.com/LiskHQ/lisk-nano/blob/development/src/components/app/index.test.js#L6

src/components/registerDelegate/index.test.js
```diff
@@ -0,0 +1,87 @@
...
+    expect(typeof props.onAccountUpdated).to.be.equal('function');
+    expect(typeof props.showSuccessAlert).to.be.equal('function');
+    expect(typeof props.showErrorAlert).to.be.equal('function');
+  });
+
+  describe('onAccountUpdated', () => {
+    it('should return a dispatch object', () => {
+      const props = mountedAccount.find(RegisterDelegate).props();
+      const data = props.onAccountUpdated(account);
+      expect(data).to.be.equal();
```
slaweet: This looks wrong. First `data` should not be `undefined`, second `data` comes from your mocked function so it's not the thing that needs to be checked. 

It would be better to `import sinonChai from 'sinon-chai'; chai.use(sinonChai);` then define  `store.onAccountUpdated = sinon.spy();`  and `expect(store.onAccountUpdated).to.have.been.calledWith();`

src/components/registerDelegate/index.test.js
```diff
@@ -0,0 +1,87 @@
...
+    });
+  });
+
+  describe('showSuccessAlert', () => {
+    it('should return a dispatch object', () => {
+      const props = mountedAccount.find(RegisterDelegate).props();
+      const data = props.showSuccessAlert('sample text');
+      expect(data).to.be.equal();
```
slaweet: same here

src/components/registerDelegate/index.test.js
```diff
@@ -0,0 +1,87 @@
...
+    });
+  });
+
+  describe('showErrorAlert', () => {
+    it('should return a dispatch object', () => {
+      const props = mountedAccount.find(RegisterDelegate).props();
+      const data = props.showErrorAlert('sample text');
+      expect(data).to.be.equal();
```
slaweet: same here

src/components/registerDelegate/registerDelegate.test.js
```diff
@@ -0,0 +1,128 @@
...
+    delegateApiMock.verify();
+    delegateApiMock.restore();
+  });
+
+  describe('Ordinary account', () => {
+    beforeEach(() => {
+      wrapper = mount(<RegisterDelegate {...normalProps} />);
+    });
+
+    it('renders an InfoParagraph components', () => {
```
slaweet: typo "components" -> "component"

src/components/registerDelegate/registerDelegate.test.js
```diff
@@ -0,0 +1,128 @@
...
+      name: 'Custom Node',
+      custom: true,
+      address: 'http://localhost:4000',
+      testnet: true,
+      nethash: '198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d',
+    }),
+  },
+  closeDialog: () => {},
+  onAccountUpdated: () => {},
+  showSuccessAlert: () => {},
```
slaweet: You can make this `showSuccessAlert: sinon.spy()` and then expect that it has been called when necessary. 

Those tests ending with `wrapper.find('.submit-button').simulate('click');` and not checking anything with `expect`.

src/components/registerDelegate/registerDelegate.test.js
```diff
@@ -0,0 +1,128 @@
...
+  showErrorAlert: () => {},
```
slaweet: same here

src/components/registerDelegate/registerDelegate.test.js
slaweet: I noticed that I'm not doing this either (in `send.test.js`). I tried to but the spy does not seem to be called.

src/components/registerDelegate/registerDelegate.js
```diff
@@ -0,0 +1,77 @@
...
+          autoFocus={true}
+          className='username'
+          onChange={this.changeHandler.bind(this, 'name')}
+          error={this.state.nameError}
+          value={this.state.name} />
+          {
+             this.props.account.secondSecret &&
+              <Input label='Second secret' required={true}
+                className='second-secret'
+                value={this.state.secondSecret} />
```
slaweet: The `Input` won't set `this.state.secondSecret` without `onChange`

slaweet:
Now that pending transactions PR is merged this should also dispatch a pending transaction (see how it's used in send send https://github.com/LiskHQ/lisk-nano/blob/development/src/components/send/send.js#L74-L81 )

PR #544, opened 2017-08-01
Extract css styles to one file - Closes #511

### LINKED ISSUE #511: Page is rendered before styles are loaded
### Expected behaviour
Page should render only with the styles applied.
### Actual behaviour
When loading the app there is a few miliseconds when the page is rendered, but the styles are not applied. e.g. there is is a huge logo.
### Steps to reproduce
- Go to http://localhost:8080/ 
- Refresh the page

alepop:
close #511

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "redux-logger": "=3.0.6"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


slaweet (approved):
Looks good.

PR #549, opened 2017-08-03
Implement pending transactions in React - Closes #514

### LINKED ISSUE #514: Implement pending transactions in React
### Expected behaviour
Transactions tab should show pending transactions. Send dialog should store sent transactions to redux. Other dialogs that create a transaction (vote, register 2nd passphrase, register delegate) should also use this.
### Actual behaviour
There are no pending transactions

slaweet:
Closes #514


src/store/reducers/transactions.js
```diff
@@ -0,0 +1,40 @@
...
+      });
+    case actionTypes.transactionsLoaded:
+      return Object.assign({}, state, {
+        confirmed: [
+          ...state.confirmed,
+          ...action.data,
+        ],
+      });
+    case actionTypes.transactionsUpdated:
+      startTimesamp = state && state.confirmed && state.confirmed.length ?
```
reyraa: Having in mind that we've defined the initial state of `state` equal to `{ pending: [], confirmed: [] }`. I think we just need to check `state.confirmed.length`.

src/store/reducers/transactions.js
```diff
@@ -0,0 +1,40 @@
...
+        state.confirmed[0].timestamp :
+        0;
+      return Object.assign({}, state, {
+        pending: state.pending.filter(
+          tx => action.data.filter(tx2 => tx2.id === tx.id).length === 0),
```
reyraa: Please use a more descriptive name instead of `tx2`.

src/store/reducers/transactions.js
```diff
@@ -0,0 +1,40 @@
...
+        confirmed: [
+          ...action.data.filter(tx => tx.timestamp > startTimesamp),
```
reyraa: Since we don't need to perform the complex test in line 24, we can move the logic to this line and get rid of the variable `startTimesamp `.

src/components/transactions/transactionsComponent.js
```diff
@@ -9,33 +9,18 @@ class Transactions extends React.Component {
...
-  componentDidMount() {
-    this.loadMore();
-  }
-
   loadMore() {
-    if (this.state.loadMore && this.state.length > this.state.offset) {
-      this.setState({ loadMore: false });
-      transactions(this.props.activePeer, this.props.address, 20, this.state.offset)
+    if (this.state.canLoadMore) {
+      this.setState({ canLoadMore: false });
```
reyraa: As mentioned in official documents:

> After the initial mounting of components, a re-rendering will occur when:
> A component's setState() method is called
> A component's forceUpdate() method is called

So it's a good practice not to use `setState` if you don't intend to re-render the component. In this case, I'll suggest using `this. canLoadMore` instead of `this.state. canLoadMore`.

src/store/reducers/transactions.js
```diff
@@ -0,0 +1,40 @@
+import actionTypes from '../../constants/actions';
+
+/**
+ *
+ * @param {Array} state
+ * @param {Object} action
+ */
+const dialog = (state = { pending: [], confirmed: [] }, action) => {
```
reyraa: Please rename `dialog` to `transactions`.

src/store/reducers/transactions.js
reyraa: And please drop a line of comment here to make the logic easier to understand. thanks

src/components/spinner/spinner.css
```diff
@@ -0,0 +1,41 @@
...
+  -webkit-animation-delay: -0.32s;
+  animation-delay: -0.32s;
+}
+.bounce2 {
+  -webkit-animation-delay: -0.16s;
+  animation-delay: -0.16s;
+}
+
+// Spinner animations
+@-webkit-keyframes sk-bouncedelay {
```
reyraa: `postcss` generates vendor prefixes, no need to add them manually. and comments with double slash (`//`) cause the next rule fail. please use `/**/`
You might like o read [this](https://css-tricks.com/snippets/css/comments-in-css/).

src/store/reducers/transactions.js
```diff
@@ -0,0 +1,43 @@
...
+      });
+    case actionTypes.transactionsLoaded:
+      return Object.assign({}, state, {
+        confirmed: [
+          ...state.confirmed,
+          ...action.data,
+        ],
+      });
+    case actionTypes.transactionsUpdated:
+      startTimesamp = state.confirmed.length ?
```
reyraa: Typo in the name of variable. `startTimesamp` > `startTimestamp `

PR #552, opened 2017-08-03
Set second passphrase - Closes #353

### LINKED ISSUE #353: Migrate Register second passphrase to React

reyraa:
- Uses PricedButton in ActionBar, therefore it needs to adapt the unit tests in many components.
 - Creates a button triggering a  modal to generate and register a second passphrase to current account.
 - Also includes some minor eslint fixings on other components.

Closes #353


src/components/secondPassphrase/index.js
```diff
@@ -0,0 +1,52 @@
...
+    !props.account.secondSignature ?
+      <MenuItem caption="Register second passphrase"
+        onClick={() => props.setActiveDialog({
+          title: 'Second Passphrase',
+          childComponent: Passphrase,
+          childComponentProps: {
+            onPassGenerated: onLoginSubmission,
+            keepModal: true,
+          },
+        })}/> : <li className={`empty-template ${styles.hidden}`}></li>
```
slaweet: Why not just `})}/> : null`?

src/components/secondPassphrase/secondPassphrase.css
```diff
@@ -0,0 +1,3 @@
+.hidden {
+  display: none;
+}
```
slaweet: This file can be deleted after the resolving comment on line that uses it.

src/components/secondPassphrase/index.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const withSecondPassAccount = {
+    isDelegate: true,
+    address: '16313739661670634666L',
+    balance: 1000e8,
+    secondSignature: 'sample phrase',
+  };
+
+  it('renders one MenuItem component for a normal account', () => {
+    const props = {
```
slaweet: To prevent code duplication, `props` can be defined outside `it`s without `account`. Appropriate `account` can be set inside each `it`.

src/components/secondPassphrase/index.js
reyraa: because a component can not return null. it's possible when you're doing this in the middle of a template when at the end, the component won't return null, but not here.

src/components/secondPassphrase/secondPassphrase.css
reyraa: since as I mentioned above, I'll have to use it, then this comment also needs to saty

reyraa:
yes, I've also seen the issue with missing word. there are other parts I want to tweak in passphrase component. I'll create a ticket fix them up all together.

src/components/secondPassphrase/index.js
slaweet: It can return `null`
https://stackoverflow.com/questions/42083181/is-it-possible-to-return-empty-in-react-render-function
I just tested it.

PR #560, opened 2017-08-04
Create a React component for second passphrase field - Closes #540

### LINKED ISSUE #540: Create a React component for second passphrase field
Send dialog in React currently doesn't support second passphrase accounts. It should display the second passphrase field if the account has it. There should be a unified solution to use across all transaction types (send, vote, register delegate).

slaweet:
Closes #540

### DIFF
--- a/src/components/secondPassphraseInput/index.test.js
+++ b/src/components/secondPassphraseInput/index.test.js
@@ -3,11 +3,10 @@ import chai, { expect } from 'chai';
 import sinonChai from 'sinon-chai';
 import { mount } from 'enzyme';
 import { Provider } from 'react-redux';
-import SecondPassphraseInputContainer from './';
+import SecondPassphraseInputContainer from './index';
 import store from '../../store';
 
 chai.use(sinonChai);
-
 
 describe('SecondPassphraseInputContainer', () => {
   let wrapper;
--- a/src/components/secondPassphraseInput/index.test.js
+++ b/src/components/secondPassphraseInput/index.test.js
@@ -3,19 +3,28 @@ import chai, { expect } from 'chai';
 import sinonChai from 'sinon-chai';
 import { mount } from 'enzyme';
 import { Provider } from 'react-redux';
+import configureMockStore from 'redux-mock-store';
 import SecondPassphraseInputContainer from './index';
-import store from '../../store';
 
 chai.use(sinonChai);
 
 describe('SecondPassphraseInputContainer', () => {
   let wrapper;
 
-  beforeEach(() => {
-    wrapper = mount(<Provider store={store}><SecondPassphraseInputContainer /></Provider>);
+  it('should render SecondPassphraseInput with props.hasSecondPassphrase if store.account.secondSignature is truthy', () => {
+    const store = configureMockStore([])({ account: { secondSignature: 1 } });
+    wrapper = mount(<Provider store={store}>
+        <SecondPassphraseInputContainer onError={ () => {} } />
+      </Provider>);
+    expect(wrapper.find('SecondPassphraseInput')).to.have.lengthOf(1);
+    expect(wrapper.find('SecondPassphraseInput').props().hasSecondPassphrase).to.equal(true);
   });
 
-  it('should render SecondPassphraseInput', () => {
-    expect(wrapper.find('SecondPassphraseInput')).to.have.lengthOf(1);
+  it('should render SecondPassphraseInput with !props.hasSecondPassphrase if store.account.secondSignature is falsy', () => {
+    const store = configureMockStore([])({ account: { secondSignature: 0 } });
+    wrapper = mount(<Provider store={store}>
+        <SecondPassphraseInputContainer onError={ () => {} } />
+      </Provider>);
+    expect(wrapper.find('SecondPassphraseInput').props().hasSecondPassphrase).to.equal(false);
   });
 });
--- a/src/components/secondPassphraseInput/index.test.js
+++ b/src/components/secondPassphraseInput/index.test.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import chai, { expect } from 'chai';
+import sinonChai from 'sinon-chai';
+import { mount } from 'enzyme';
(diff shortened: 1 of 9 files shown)


src/components/secondPassphraseInput/secondPassphraseInput.js
```diff
@@ -0,0 +1,42 @@
...
+  handleChange(value) {
+    this.props.onChange(value);
+    const error = this.validate(value);
+    if (error) {
+      this.props.onError(error, value);
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  validate(value) {
```
yasharAyari: If you don't use 'this' in a function of class, it doesn't need to be a function inside a class. So,you should define it outside your class and use it inside the class functions. after that you remove 'eslint-disable-next-line'.

slaweet: This `eslint` check is IMO stupid. My reason was that the function logically belongs to the component. There is a very similar `validateInput` function in send component [0], which by chance needs to use `this` so `eslint` is not complaining. But both functions are from the semantic point of view the same (validators) and thus both should IMO be also syntactically the same. 

[0] https://github.com/LiskHQ/lisk-nano/blob/development/src/components/send/send.js#L51-L62

src/components/secondPassphraseInput/index.test.js
```diff
@@ -0,0 +1,22 @@
...
+import sinonChai from 'sinon-chai';
+import { mount } from 'enzyme';
+import { Provider } from 'react-redux';
+import SecondPassphraseInputContainer from './';
+import store from '../../store';
+
+chai.use(sinonChai);
+
+
+describe('SecondPassphraseInputContainer', () => {
```
reyraa: Please cover the goal of component.

src/components/secondPassphraseInput/index.test.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import chai, { expect } from 'chai';
+import sinonChai from 'sinon-chai';
+import { mount } from 'enzyme';
+import { Provider } from 'react-redux';
+import SecondPassphraseInputContainer from './';
```
reyraa: Please use the full name of file without postfix.

src/components/secondPassphraseInput/secondPassphraseInput.js
```diff
@@ -0,0 +1,42 @@
...
+import { isValidPassphrase } from '../../utils/passphrase';
+
+class SecondPassphraseInput extends React.Component {
+  componentDidMount() {
+    if (this.props.hasSecondPassphrase) {
+      this.props.onError(undefined, '');
+    }
+  }
+
+  handleChange(value) {
```
reyraa: This is not a generic change handler. Please use a more descriptive name.

src/components/secondPassphraseInput/secondPassphraseInput.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  validate(value) {
+    if (!value) {
+      return 'Required';
+    } else if (!isValidPassphrase(value)) {
+      return 'Invalid passphrase';
+    }
+    return undefined;
```
reyraa: This is redundant. All void functions return undefined.

reyraa: In case eslint complains, define a variable and return once.

src/components/secondPassphraseInput/secondPassphraseInput.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  render() {
+    return (this.props.hasSecondPassphrase ?
+      <Input label='Second Passphrase' required={true}
+        className='second-passphrase'
+        error={this.props.error}
+        value={this.props.value}
+        onChange={this.handleChange.bind(this)} /> :
+      null);
```
reyraa: A component shouldn't return null as template. What happens if we use this component for a account without a second passphrase?

src/components/secondPassphraseInput/index.test.js
```diff
@@ -0,0 +1,22 @@
...
+import store from '../../store';
+
+chai.use(sinonChai);
+
+
```
reyraa: Please remove extra blank line.

src/components/secondPassphraseInput/secondPassphraseInput.test.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import chai, { expect } from 'chai';
+import sinonChai from 'sinon-chai';
+import { mount } from 'enzyme';
+import sinon from 'sinon';
+import SecondPassphraseInput from './secondPassphraseInput';
+
+chai.use(sinonChai);
+
+
```
reyraa: Please remove the extra blank line

src/components/secondPassphraseInput/secondPassphraseInput.js
slaweet: It's ok: https://facebook.github.io/react/docs/conditional-rendering.html#preventing-component-from-rendering

src/components/secondPassphraseInput/index.test.js
slaweet: If you think we shouldn't use more than one line, please open an Issue to update eslint http://eslint.org/docs/rules/no-multiple-empty-lines

src/components/secondPassphraseInput/secondPassphraseInput.js
slaweet: I know it's redundant. But AFAIK that's the whole point of `consistent-return` eslint rule. To improve readability through redundancy. I think it's also more readable than using a variable and return once.

reyraa: `return undefined` might help eslint but it's not a logic to cover consistent return pattern. as mentioned before, you can define a variable and assign instead, then . return just once.

Issue #548, opened 2017-08-02
Jenkins doesn't fail if there are eslint errors in tests


slaweet:
### Expected behaviour
Jenkins should fail if there are some eslint errors in any js file
That includes:
- unit tests
- e2e tests.
- electron js files.

### Actual behaviour
Jenkins only fails of there are eslint errors in source code, not in tests, so these errors made it to `development`:
```
ERROR in ./src/actions/toaster.test.js
ESLintError: 
/home/lisk/jenkins/workspace/Nano-Pipeline_PR-528-FIT557NSAL6UDXDBXECWKK7CZ7AOPHW3SVG5CBQI5YI6GO5HABMA/src/actions/toaster.test.js
  11:55  error  Block must not be padded by blank lines  padded-blocks

��� 1 problem (1 error, 0 warnings)


ERROR in ./src/components/passphrase/passphraseGenerator.test.js
ESLintError: 
/home/lisk/jenkins/workspace/Nano-Pipeline_PR-528-FIT557NSAL6UDXDBXECWKK7CZ7AOPHW3SVG5CBQI5YI6GO5HABMA/src/components/passphrase/passphraseGenerator.test.js
  5:10  error  'mount' is defined but never used  no-unused-vars

��� 1 problem (1 error, 0 warnings)


ERROR in ./src/utils/passphrase.test.js
ESLintError: 
/home/lisk/jenkins/workspace/Nano-Pipeline_PR-528-FIT557NSAL6UDXDBXECWKK7CZ7AOPHW3SVG5CBQI5YI6GO5HABMA/src/utils/passphrase.test.js
  2:10  error  'spy' is defined but never used  no-unused-vars

��� 1 problem (1 error, 0 warnings)


ERROR in ./src/components/voting/votingHeader.test.js
ESLintError: 
/home/lisk/jenkins/workspace/Nano-Pipeline_PR-528-FIT557NSAL6UDXDBXECWKK7CZ7AOPHW3SVG5CBQI5YI6GO5HABMA/src/components/voting/votingHeader.test.js
  44:55  error  Missing semicolon  semi

��� 1 problem (1 error, 0 warnings)
```

PR #577, opened 2017-08-08
Re-enable e2e tests for features already working in React - Closes #565

### LINKED ISSUE #565: Re-enable e2e tests for features already working in React
### Expected behaviour
All currently migrated behaviour should be covered with e2e tests. 
### Actual behaviour
Some behaviour is not  covered - e.g. Send

slaweet:
Closes #565


reyraa:
There's 3 thoughts about current setup of e2e tests:
 - They don't meet out eslint standard. which we'll cover with #548 
 - It depicts active peer as Offline in all the tests. even though it makes all the Api calls successfully.
 - we always have to fun `e2e-test-setup.sh` before running `e2e-test`, why not merging these together?
 - Some of tests are written based on  classnames belonging to the theme framework and thus vulnerable to theme changes.

PR #582, opened 2017-08-08
Unit tests should run on file change only once - Closes #581

### LINKED ISSUE #581: Unit tests should run on file change only once
### Expected behaviour
When running `npm run test-live` tests should run only once on file change.
### Actual behaviour
When running `npm run test-live` tests run twice on file change.

It's probably related to this [0]
> Actually, watching with webpack is better since the tests will be rerun only once - after webpack rebuilds test_index.js.
> Watching with karma means that karma will run tests upon change some_test.js and after webpack rebuilds the test_index.js

[0] https://github.com/webpack-contrib/karma-webpack/issues/23#issuecomment-55240090

alepop:
closes #581

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -14,7 +14,7 @@ module.exports = function (config) {
     frameworks: ['mocha', 'chai'],
     files: [
       fileRoot,
-      { pattern: filePattern, included: false, served: false, watched: true }
+      { pattern: filePattern, included: false, served: false, watched: false },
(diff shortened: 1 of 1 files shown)


slaweet (approved):
That's great @alepop.
I would swear that I tried the same thing and it didn't work, but your's works :-)

alepop:
@slaweet magic! :-)

PR #609, opened 2017-08-16
Migrate desktop notifications to react - Closes #568

### LINKED ISSUE #568: Migrate desktop notifications for changes to React
### Expected behaviour
The goal is to migrate the functionality implemented in Angular the original issue:
https://github.com/LiskHQ/lisk-nano/issues/253
### Actual behaviour
The functionality is not in React version, yet.

alepop:
closes #568


alepop:
@slaweet I merged the branch with development. I checked it and HTTP requests are triggered when the app is not in focus and notification was triggered for me.

src/store/middlewares/notification.js
```diff
@@ -0,0 +1,23 @@
+import actionTypes from '../../constants/actions';
+import Notification from '../../utils/notification';
+
+const notificationMiddleware = (store) => {
+  const notify = Notification.init();
+  return next => (action) => {
+    next(action);
+
+    const { account } = store.getState();
```
slaweet: So I figured out that the reason it wasn't showing the notification is actually here. Lines 7 and 9 need to be swapped. 

Currently, `next(action);` goes synchronously through all other middlewares and reducers, then `{ account }` gets the updated value of the account, so there is never any difference in balance.

alepop: oh... you right. i handled notification with hard-coded value for testing after your first comment.

PR #613, opened 2017-08-17
Cetralise test setup - Closes #589

### LINKED ISSUE #589: Centralise test setup
### Expected behaviour

Chai plugins etc should be initialised once in a setup file for all tests.

### Actual behaviour

They're imported and initialised in individual files many times over.

slaweet:
Closes #589


src/tests.js
```diff
@@ -1,3 +1,15 @@
...
+import sinon from 'sinon';
+import sinonStubPromise from 'sinon-stub-promise';
+
+chai.use(sinonChai);
+chai.use(chaiEnzyme());
+chai.use(chaiAsPromised);
+sinonStubPromise(sinon);
+
 // load all tests into one bundle
 const testsContext = require.context('.', true, /\.test\.js$/);
```
yasharAyari: Use es6 import instead of require.

slaweet: I don't think it is possible to use `import` for this task. If you know how, please tell me.

yasharAyari: you can try this https://github.com/crisberrios/dir-loader

slaweet: I looked at `dir-loader`, but it is still using `require`, so I don't see how does it help. 

I don't see what is the problem with the current solution.

PR #617, opened 2017-08-18
Fix responsiveness of the layout - Closes #529

### LINKED ISSUE #529: Fix responsiveness of the layout
### Expected behaviour

responsive-ok

Compare `develpment` with `1.0.0` on different window widths to see what else needs to be done.
### Actual behaviour
responsive-nok

slaweet:
Closes #529

### DIFF
--- a/src/components/account/account.css
+++ b/src/components/account/account.css
@@ -69,6 +69,13 @@
     font-weight: 500;
     letter-spacing: 0;
     margin-top: 0;
+    margin-top: 20px;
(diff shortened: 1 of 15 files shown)


reyraa:
I think there's enough space for menu to stay in the right side of logo:
screen shot 2017-08-21 at 10 18 12

I think transaction ID is taking more space and is less informative than the others. if we hide it in this media query, we'll have a better user experience:
screen shot 2017-08-21 at 10 18 39

PR #630, opened 2017-08-21
Fix issues in forging - Closes #618

### LINKED ISSUE #618: Issues in forging tab
### Expected behaviour
forging tab should work as before the migration
### Actual behaviour
forging tab
- shows Rank higher by 1 (compared to nano 1.0.0)
- doesn’t round LSK amounts to 2 decimal places
- Forged blocks don’t get updates after the initial load

slaweet:
Closes #618

### DIFF
--- a/src/store/middlewares/account.js
+++ b/src/store/middlewares/account.js
@@ -3,6 +3,7 @@ import { accountUpdated } from '../../actions/account';
 import { transactionsUpdated } from '../../actions/transactions';
 import { activePeerUpdate } from '../../actions/peers';
 import actionTypes from '../../constants/actions';
+import { fetchAndUpdateForgedBlocks } from '../../actions/forging';
 
 const updateAccountData = next => (store) => { // eslint-disable-line
   const { peers, account } = store.getState();
@@ -15,6 +16,14 @@ const updateAccountData = next => (store) => { // eslint-disable-line
         confirmed: response.transactions,
         count: parseInt(response.count, 10),
       })));
+      if (account.isDelegate) {
+        store.dispatch(fetchAndUpdateForgedBlocks({
+          activePeer: peers.data,
+          limit: 10,
+          offset: 0,
+          generatorPublicKey: account.publicKey,
+        }));
+      }
     }
     next(accountUpdated(result));
   });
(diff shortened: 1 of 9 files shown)


reyraa:
The previous version doesn't show Forged blocks header if the list is empty. instead it shows a message: You have not forged any blocks yet. I think it makes sense to keep the same behaviour.

reyraa:
I can find how have you fixed this issue:
> shows Rank higher by 1

slaweet:
"Rank higher by 1" was actually a slightly different problem.
I was testing on delegate with rank 50 and it showed 51. But later when I tried rank 49, it showed 52. So it was subtracting from 101, the same way it does for the progress bar. The fix is this:
https://github.com/LiskHQ/lisk-nano/pull/630/commits/9247adff39cff49cbdff3df4787cbbac9d4dcf76

src/store/middlewares/account.test.js
```diff
@@ -40,5 +47,41 @@ describe('Account middleware', () => {
...
+
+  it(`should call store.dispatch(fetchAndUpdateForgedBlocks(...)) on ${actionTypes.metronomeBeat} action if account.balance changes and account.isDelegate`, () => {
+    state.account.isDelegate = true;
+    store.getState = () => (state);
+    const stubGetAccount = stub(accountApi, 'getAccount').resolves({ balance: 10e8 });
+    const stubGetAccountStatus = stub(accountApi, 'getAccountStatus').resolves(true);
+
+    middleware(store)(next)({ type: actionTypes.metronomeBeat });
+
+    expect(stubGetAccount).to.have.been.calledWith();
```
reyraa: This test is not related to title. 
The issue is, Actions that we create using Redux thunk don't return object. They return a function which accepts `dispatch` as argument. but you're using just like other actions.

Other than how you've used it, the actual issue is as described above it best called inside component class.

src/store/middlewares/account.js
```diff
@@ -15,6 +16,14 @@ const updateAccountData = next => (store) => { // eslint-disable-line
         confirmed: response.transactions,
         count: parseInt(response.count, 10),
       })));
+      if (account.isDelegate) {
+        store.dispatch(fetchAndUpdateForgedBlocks({
```
reyraa: This action returns a function that accepts dispatch as argument. Prefer using it inside the component when account property is updated and shows a change in balance.

PR #631, opened 2017-08-21
Display correct rank rate

alepop:
**Before**:
screen shot 2017-08-21 at 17 32 11
**After**:
screen shot 2017-08-21 at 17 31 24


slaweet:
Thanks, but I was fixing the same issue today:
https://github.com/LiskHQ/lisk-nano/pull/630/commits/9247adff39cff49cbdff3df4787cbbac9d4dcf76
And I have it in a PR with more fixes of Forging tab, so I'll selfishly prefer my PR.

alepop:
@slaweet ok, np

PR #636, opened 2017-08-22
Isolate core and presentational logics - Closes #611, #610, #584

### LINKED ISSUE #584: Logout should remove all account-related data from redux
### Expected behaviour
Logout should remove all account-related data from redux. E.g. transactions, forging, etc.
Alternatively, it can be taken care of in the respective components.
### Actual behaviour
Logout leaves data in redux, so on next login transactions list displays transactions from the previous account.
### Steps to reproduce
- Login as account A
- Logout
- Login as account B

### LINKED ISSUE #610: Create a middleware to handle success alerts after a transaction added
After every successful transaction of any type, we'll show a success alert notifying user about the Api call status. This can be handled in a middleware in order to keep up with declarative programming paradigm and organise our messages.

### LINKED ISSUE #611: Move Api calls to actions using Redux Thunk
### Expected behaviour
To keep our react component pure presentational, we have to move our Api calls to actions. 
With this change, we can distribute the logic between multiple actions and middlewares. In addition to the above mentioned reason, this improvement will help us test our logic in pure JS instead of traversing and manipulating DOM elements.

### Actual behaviour
Currently almost all of them are implemented inside the components.

reyraa:
- Creates a middleware to:
   - Gather account info
   - Make Api call and handle success/error situations.
   - dispatch new actions accordingly.
 - Removes all account information from store after logging out
 - Adds Redux Thunk:
   -  Moves all Api calls from components leaving only a single action in them to be fired when required.
   - Covers all actions unit tests.
 - Creates the TransactionAdded middleware to centralise handling of pending transactions and subsequent alert dialog.

Closes #611
Closes #610
Closes #584


src/actions/account.js
```diff
@@ -18,9 +26,13 @@ export const accountUpdated = data => ({
...
-export const accountLoggedOut = () => ({
-  type: actionTypes.accountLoggedOut,
-});
+export const accountLoggedOut = () =>
+  (dispatch) => {
+    dispatch(forgingReset());
+    dispatch(activePeerReset());
+    dispatch(transactionsReset());
+    dispatch({ type: actionTypes.accountLoggedOut });
+  };
```
alepop: Each dispatch will handle view re-render. I recommend to using one action that will be handled in different reduces. like, handle `logout` in each of it.

src/components/forging/forging.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Card } from 'react-toolbox/lib/card';
+import Waypoint from 'react-waypoint';
+import ForgingTitle from './forgingTitle';
+import DelegateStats from './delegateStats';
+import ForgingStats from './forgingStats';
+import ForgedBlocks from './forgedBlocks';
+
+class Forging extends React.Component {
+  loadStats(key, startMoment) {
```
alepop: Component do not have a state, better use the stateless one

src/actions/account.js
reyraa: Good point. I'll change it

src/components/forging/forging.js
reyraa: yeah I forgot to change it

PR #643, opened 2017-08-23
Fix infinity scroll - Closes #623

### LINKED ISSUE #623: Infinite scroll loading should start earlier than when the bottom is hit
### Expected behaviour
Infinite scroll loading should start earlier than when the bottom is hit (~1 screen height earlier).
It worked like that before React migration and felt more smooth.
### Actual behaviour
Infinite scroll loading starts when bottom of the page is hit
### Steps to reproduce
- Open Nano
- Login
- Go to "delegates"
- scroll

alepop:
closes #623

### DIFF
--- a/features/transactions.feature
+++ b/features/transactions.feature
@@ -2,9 +2,9 @@ Feature: Transactions tab
   Scenario: should show transactions
     Given I'm logged in as "genesis"
     When I click tab number 1
-    Then I should see table with 20 lines
+    Then I should see table with 40 lines
 
-  Scenario: should allow send to address 
+  Scenario: should allow send to address
     Given I'm logged in as "genesis"
     When I click tab number 1
     And I click "from-to" element on table row no. 1
--- a/src/components/forging/forging.js
+++ b/src/components/forging/forging.js
@@ -27,6 +27,7 @@ const Forging = ({
     });
   };
 
+
   return (
     <Card style={{ padding: 8 }}>
       {account && account.isDelegate ?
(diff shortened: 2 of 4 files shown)


slaweet (changes requested):
Thanks for the PR, it's a good start.
I can see two issues:
- the infinite scroll in transactionsComponent.js doesn't work. It's because of recent refactoring, not because of this PR, but it makes sense to fix it here. The problem is not setting `this.canLoadMore = true` after the request is finished.
- The scrolling still doesn't feel smooth - I'm often waiting on the end of the list for the finish of loading. Sometimes I even had to scroll back a little to load more. Tested on mainnet list of delegates.

reyraa:
Please consider adding `scrollableAncestor` to make sure the scroll event listener is bound to window:
```
<Waypoint
          scrollableAncestor={window}
          onEnter={this.loadMore.bind(this)}>
</Waypoint>
```
this will solve the issue.

alepop:
@reyraa thanx 👍

slaweet:
It still doesn't load more transactions with some screen heights. For example, I was running with `1343px`. I think that the problem is that the waypoint is so high that it is unable to reach the bottom.

slaweet:
Now it works good, just this one e2e test needs to be updated:
```
Failures:

1) Scenario: should show transactions - features/transactions.feature:2
   Step: Then I should see table with 20 lines - features/transactions.feature:5
   Step Definition: features/step_definitions/generic.step.js:74
   Message:
     AssertionError
         + expected - actual

         -40
         +20
```
https://jenkins.lisk.io/job/lisk-nano/job/PR-643/4/console

slaweet (approved):

PR #644, opened 2017-08-24
Delegate list gets slow when loading ~1000 delegates - Closes #260

### LINKED ISSUE #260: Delegate list gets slow when loading ~1000 delegates

slaweet:
Closes #260

### DIFF
--- a/src/components/voting/votingRow.js
+++ b/src/components/voting/votingRow.js
@@ -12,9 +12,15 @@ const setRowClass = ({ pending, selected, voted }) => {
   return voted ? styles.downVoteRow : '';
 };
 
-const VotingRow = (props) => {
-  const { data } = props;
-  return (<TableRow {...props} className={`${styles.row} ${setRowClass(data)}`}>
+class VotingRow extends React.Component {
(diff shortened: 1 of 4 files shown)


src/components/voting/votingRow.js
```diff
@@ -12,9 +12,15 @@ const setRowClass = ({ pending, selected, voted }) => {
   return voted ? styles.downVoteRow : '';
 };
 
-const VotingRow = (props) => {
-  const { data } = props;
-  return (<TableRow {...props} className={`${styles.row} ${setRowClass(data)}`}>
+class VotingRow extends React.Component {
+  shouldComponentUpdate(nextProps) {
```
yasharAyari: This line cause an eslint error.

PR #651, opened 2017-08-24
Fix voting tab after migration differences - Close #622

### LINKED ISSUE #622: Fix voting tab after-migration differences
### Expected behaviour
Voting tab should work like before the migration.
### Actual behaviour
There are these differences:
- Confirm dialog doesn’t have horizontal lines around the info text
- “My votes” button in voting is not disabled when 0 votes

yasharAyari:
closes [#622](https://github.com/LiskHQ/lisk-nano/issues/622)

### DIFF
--- a/src/components/voting/voting.css
+++ b/src/components/voting/voting.css
@@ -50,6 +50,9 @@
         line-height: 24px;
         margin-left: 6px;
     }
+    & :global .disable{
+        color: rgba(0,0,0,0.38);
+    }
 }
 .voted {
     color: #7cb342;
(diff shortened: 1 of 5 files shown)


src/components/voting/voting.css
```diff
@@ -50,6 +50,9 @@
         line-height: 24px;
         margin-left: 6px;
     }
+    & :global .disable{
```
slaweet: Why `:global`? Cannot it be used as `styles.disable` instead of `'disable'`?

src/components/voting/voting.css
```diff
@@ -50,6 +50,9 @@
...
+        color: rgba(0,0,0,0.38);
```
slaweet: `.disable` also needs ` cursor: default;`

src/components/voting/voting.css
```diff
@@ -69,9 +72,18 @@
...
+    margin: 10px -24px 20px;
+}
+.autoCompleteTile{
+    margin-bottom: 5px;
+}
 /* react toolbar overwroght */
 .input{
-    margin-top: -15px;
+    margin-top: -10px;
+    padding-bottom: 5px;
```
slaweet: IMO it looks better without this `padding-bottom` override. (I like more the default of 20px)

PR #657, opened 2017-08-26
Fix pluralization for confirmation tooltip

ccampbell:
This is the tiniest of tiny issues, but I was checking a transaction earlier and noticed that it said

> 1 confirmations 

Attention to little details is important 😄


slaweet (approved):
Thanks @ccampbell

PR #673, opened 2017-08-28
Fix login page after-migration differences - Closes #624

### LINKED ISSUE #624: Fix login page after-migration differences
### Expected behaviour
Login should work like before the migration.
### Actual behaviour
There are these differences:
- Allows to log in to an offline custom node
- Doesn’t have the “Check for new version” error if API responds with 404
- Some elements are centered instead of left-aligned: Items in network dropdown, error messages on address and passphrase inputs
- Marks URLs with default ports as "Invalid", e.g. http://node03.lisk.io:80 or https://node03.lisk.io:443 Also if the ports are not specified at all.

alepop:
- [x] Prevent to log in to an offline node
- [x] Add message for error API response with 404
- [x] Left-align elements on login page
- [x] Marks URLs with default ports as valid
closes #624

### DIFF
--- a/src/store/middlewares/offline.js
+++ b/src/store/middlewares/offline.js
@@ -2,29 +2,29 @@ import actionsType from '../../constants/actions';
 import { successToastDisplayed, errorToastDisplayed } from '../../actions/toaster';
 import { loadingStarted, loadingFinished } from '../../utils/loading';
 
-const getErrorMessage = (errorCode, optData) => {
+const getErrorMessage = (errorCode, address) => {
+  let message = `Failed to connect to node ${address}`;
   switch (errorCode) {
     case 'EUNAVAILABLE':
-      return `Failed to connect to node ${optData}`;
+      message = `Failed to connect: Node ${address} is not active`;
+      break;
     case 'EPARSE':
-      return 'Make sure that you are using the latest version of Lisk Nano.';
-    default: return '';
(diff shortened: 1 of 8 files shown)


src/store/middlewares/offline.js
```diff
@@ -2,14 +2,29 @@ import actionsType from '../../constants/actions';
 import { successToastDisplayed, errorToastDisplayed } from '../../actions/toaster';
 import { loadingStarted, loadingFinished } from '../../utils/loading';
 
+const getErrorMessage = (errorCode, optData) => {
+  switch (errorCode) {
+    case 'EUNAVAILABLE':
+      return `Failed to connect to node ${optData}`;
+    case 'EPARSE':
+      return 'Make sure that you are using the latest version of Lisk Nano.';
+    default: return '';
```
slaweet: There are some differences to the pre-migration logic of this
- The default error message used to be also `Failed to connect to node ${address}. `
- the `EUNAVAILABLE` error message is `Failed to connect: Node ${address} is not active`, not `Failed to connect to node ${address}. `
- the `EPARSE` error message is appended to the default (notice `+=`)

https://github.com/LiskHQ/lisk-nano/blob/bb27b7470cc739390e6f9744e0e46d705272d9a7/src/services/api/peers.js#L118-L126

slaweet (changes requested):

slaweet (approved):

PR #679, opened 2017-08-29
ConfirmVotes component should check validity of second passphrase - Closes #628

### LINKED ISSUE #628: ConfirmVotes component should check validity of second passphrase before firing the action
### Expected behaviour
As title suggests, ConfirmVotes component should use `passphrase` utility to validate secondPasspharase before firing `votePlaced` action.

### Actual behaviour
no validation is implemented yet

### Steps to reproduce
Try voting with a structurally wrong second passphrase. The app will call Api.

yasharAyari:
Closes [#628](https://github.com/LiskHQ/lisk-nano/issues/628)

### DIFF
--- a/src/components/secondPassphraseInput/secondPassphraseInput.js
+++ b/src/components/secondPassphraseInput/secondPassphraseInput.js
@@ -5,21 +5,18 @@ import { isValidPassphrase } from '../../utils/passphrase';
 class SecondPassphraseInput extends React.Component {
   componentDidMount() {
     if (this.props.hasSecondPassphrase) {
-      this.props.onError(undefined, '');
+      this.props.onChange('');
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/components/secondPassphraseInput/secondPassphraseInput.js
```diff
@@ -5,21 +5,18 @@ import { isValidPassphrase } from '../../utils/passphrase';
 class SecondPassphraseInput extends React.Component {
   componentDidMount() {
     if (this.props.hasSecondPassphrase) {
-      this.props.onError(undefined, '');
+      this.props.onChange('');
```
slaweet: This should be `this.props.onChange('', '');`, so that in Send dialog it doesn't show the field in red from the beginning.

PR #685, opened 2017-08-30
Omit data attribute on props in votingRow

lohek:
Fix issue with data-* attribute on votingRow
I directly add `lodash` to package.json, maybe if you prefer, i can just add `lodash.omit`

Before
image

After
image

### DIFF
--- a/src/components/voting/votingRow.js
+++ b/src/components/voting/votingRow.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { TableRow, TableCell } from 'react-toolbox/lib/table';
-import omit from 'lodash/omit';
 import Checkbox from './voteCheckbox';
 import styles from './voting.css';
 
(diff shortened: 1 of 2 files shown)


src/components/voting/votingRow.js
```diff
@@ -21,7 +22,7 @@ class VotingRow extends React.Component {
   render() {
     const props = this.props;
     const { data } = props;
-    return (<TableRow {...props} className={`${styles.row} ${setRowClass(data)}`}>
+    return (<TableRow {...omit(props, ['data'])} className={`${styles.row} ${setRowClass(data)}`}>
```
alepop: Good 👍 But you can simply use `const {data, ...other} = props;` to omit the property and pass to component `{...other}` instead of using `lodash`

lohek:
Good catch, it's ok now

src/components/voting/votingRow.js
lohek: Good catch, it's ok now

slaweet:
@lohek thank you for the PR, however, there was another open PR which was among other things removing the whole `{...props}` from `TableRow`. So your fix is no longer needed.

PR #688, opened 2017-08-30
Bugfix: Registering delegate doesn't update UI - Closes #682

### LINKED ISSUE #682: Registering delegate doesn't update UI
### Expected behaviour
When delegate registration transaction is confirmed:
- "Register delegate" menu item should be hidden
- "Voting" tab should be displayed
### Actual behaviour
Nothing changes when delegate registration transaction is confirmed.

slaweet:
Closes #682


src/components/account/address.js
```diff
@@ -6,7 +6,7 @@ const Address = (props) => {
   const title = props.isDelegate ? 'Delegate' : 'Address';
   const content = props.isDelegate ?
     (<div>
-      <p className="inner primary">
+      <p className="inner primary delegate-name">
```
yasharAyari: please remove `delegate-name` class and use an ID instead of that. when we want to select a an element in js , it is better to use ID.

src/constants/transactionTypes.js
```diff
@@ -0,0 +1,7 @@
+export default {
```
yasharAyari: we’re creating `src/constants/transactionTypes.js` when there are some other places to use this but you only uses this in one file.So, we don't need this constant.

slaweet: I use it in three files:
```
src/components/transactions/amount.js
src/store/middlewares/delegateRegistration.js
src/store/middlewares/delegateRegistration.test.js
```

src/components/account/address.js
slaweet: I added this class for e2e tests. All e2e tests use classes for selectors. It is not possible to change just this one. We can discuss if we should change classes to IDs in e2e tests, but it is definitely outside the scope of this PR.

Issue #559, opened 2017-08-04
Add option to remember account for read-only access after app was closed


slaweet:
This is the first step towards #65 
The goal is to give the user an option to save the account for read-only access on the next launch of the app. Effectively it means to
- add some UI element to give the user this option
- save public key and network to local storage or cookies and retrieve it from there on launch
- adapt all transaction dialogs to display 'passphrase' field in this read-only mode

Issue #491, opened 2017-07-17
Give better passphrase error messages than "invalid" 


slaweet:
### Expected behaviour
We should give our users more information about why the passphrase they entered is invalid.
Namely:
- Passphrase should have twelve words, entered passphrase has X
- Word "XXXX" is not on the passphrase Word List. Most similar word on the list is "XXXY"
- Passphrase is not valid
### Actual behaviour
When users try to log in with an invalid passphrase the only error message they can get is "invalid"

slaweet:
This tool can help
https://iancoleman.github.io/bip39/

PR #690, opened 2017-08-30
Allow login with abbreviated passphrases

georeith:
Lisk Nano only accepts passwords containing at least 12 BIP39 English words.

A useful feature of BIP39 is that the first 4 letters of any word in BIP39 are guaranteed to be unique.

Many people store their passwords in devices such as Cryptosteel and only store the first 4 characters.

This PR allows the user to enter their password in abbreviated form:

`feel lonely mountain suggest team rabbit brass buffalo year scare ahead twist`

can be entered as:

`feel lone moun sugg team rabb bras buff year scar ahea twis`

Without it the user has to manually look up the matching word in the BIP39 word list against their abbreviated passphrase before entering.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# http://EditorConfig.org
+root = true
+
+[*.{js,feature}]
+charset = utf-8
+indent_style = space
+indent_size = 2
--- a/README.md
+++ b/README.md
@@ -15,14 +15,14 @@ npm run dev
 
 Open http://localhost:8080
 
-For ease of development, you can set a cookie to prefill a passphrase, e.g.:
+For ease of development, you can setItem in localStorage to prefill a passphrase, e.g.:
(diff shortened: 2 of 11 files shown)


slaweet:
We had some discussion in LiskHQ about this feature and in the end, I don't want to accept it.

It is a feature for a very small fraction of our users. It is IMO a problem of devices such as Cryptosteel and they should solve it once for all wallet applications, e.g. by providing a simple single-purpose app that does resolve the abbreviated passphrase to the original one. 

Implementing this feature in each wallet app is an unnecessary duplicated effort, and not just in this one PR. Accepting this feature means that we have to maintain it in the future, which complicates tasks that concern passphrase, e.g. https://github.com/LiskHQ/lisk-nano/issues/491, https://github.com/LiskHQ/lisk-nano/issues/559

PR #724, opened 2017-09-11
Give better passphrase error messages than "invalid" - Closes #491

### LINKED ISSUE #491: Give better passphrase error messages than "invalid"

alepop:
- Display Error about incorrect passphrase length;
- If passphrase has a word that does not exist in the dictionary, then display an error with similar word based on the "Levenshtein" algorithm. (improved).
- In other case display "Passphrase is not valid".
closes #491


src/utils/similarWord.js
```diff
@@ -0,0 +1,92 @@
+const mnemonic = require('bitcore-mnemonic');
```
slaweet: Why `require` instead of `import`? I suspect there is a reason but don't know what it is. A comment would be nice.

src/components/login/login.js
```diff
@@ -85,17 +86,33 @@ class Login extends React.Component {
...
+    }
+    const mnemonic = passphrase.trim().toLowerCase().split(' ');
+    if (mnemonic.length < 12) {
+      return `Passphrase should have 12 words, entered passphrase has ${mnemonic.length}`;
+    }
+
+    const invalidWord = mnemonic.find(word => !inDictionary(word));
+    if (invalidWord && (invalidWord.length >= 2 && invalidWord.length <= 8)) {
+      const validWord = findSimilarWord(invalidWord);
+      if (validWord) {
```
slaweet: ``Word "${invalidWord}" is not on the passphrase Word List.`` error message would be helpful when no `validWord` is found.

slaweet (changes requested):
All but two minor things look good. Thank you @alepop

src/components/login/login.js
```diff
@@ -85,17 +86,37 @@ class Login extends React.Component {
...
+    if (!passphrase) {
+      return 'Empty passphrase';
+    }
+    const mnemonic = passphrase.trim().toLowerCase().split(' ');
+    if (mnemonic.length < 12) {
+      return `Passphrase should have 12 words, entered passphrase has ${mnemonic.length}`;
+    }
+
+    const invalidWord = mnemonic.find(word => !inDictionary(word));
+    if (invalidWord && (invalidWord.length >= 2 && invalidWord.length <= 8)) {
```
slaweet: This check `(invalidWord.length >= 2 && invalidWord.length <= 8)`  should be just for the `findSimilarWord(invalidWord);`. The error message `Word "${invalidWord}" is not on the passphrase Word List.` is usefull even when the invalid word is too long or too short.

alepop: @slaweet fixed

slaweet:
Jenkins is failing on `eslint` check:
```

/home/lisk/workspace/lisk-nano_PR-724-LITUYBWAM7USPXTW3P5KVJDQJEC2774E6AEOXKMD6XGX7UOQNTNA/src/utils/similarWord.test.js
  10:22  error  Absolute imports should come before relative imports  import/first

✖ 1 problem (1 error, 0 warnings)

```

slaweet (approved):

PR #750, opened 2017-09-18
Implement routing for modals - Closes #554

### LINKED ISSUE #553: Design Launch protocol
This is the first step of https://github.com/LiskHQ/lisk-nano/issues/20. The goal is to design the scheme to launch everything that Lisk Nano allows to do, 
Especially useful cases are:
- open send dialog with prefilled values,
- open vote dialog with prefilled delegates.

### LINKED ISSUE #554: Implement routing that can open anything in the app
Implement routing based on the design created in https://github.com/LiskHQ/lisk-nano/issues/553 that will later allow to Setup the launch protocol.

reyraa:
- Defines routing for dialogs
 - Creates Relativelink component
 - Edits dialog component to work independently from its callees.
 - replaces all buttons which trigger action to open a dialog with a relative link to the corresponding route

Closes #554 
Closes #553


package.json
```diff
@@ -32,6 +32,7 @@
     "bitcore-mnemonic": "=1.1.1",
     "copy-to-clipboard": "=3.0.6",
     "flexboxgrid": "=6.3.1",
+    "history": "^4.7.2",
```
slaweet: This should use `=` instead of `^`.
We should set up this in `.npmrc` file like `lisk-js` guys did https://github.com/LiskHQ/lisk-js/pull/382

reyraa: TBH, I never liked this approach, choosing `=` we easily end up with outdated dependencies.  if there's a breaking change in one of the dependencies, I'll prefer to get notified of somehow. that's exactly the reason I clean install my node modules every once in a while.

alepop: @reyraa, @slaweet https://github.com/greenkeeperio/greenkeeper

reyraa: Sure it's a good idea, but please don't tell me to add it in this PR :-D

src/components/tabs/tabs.js
```diff
@@ -8,39 +8,33 @@ const tabs = [
...
-};
+const Tabs = ({ history, isDelegate, t }) => (
+  <ToolboxTabs index={getIndex(history)}
+    theme={styles}
+    onChange={navigate.bind(null, history)}
+    className={`${styles.tabs} main-tabs`}>
+    {getTabs(isDelegate).map((tab, index) =>
+      <Tab
+        key={index}
+        label={t(tab)}
```
slaweet: Calling `t(tab)` doesn't allow i18n-scanner to find tab names for the source*.json file.

src/components/transactions/transactionType.js
```diff
@@ -1,42 +1,42 @@
...
       break;
     default:
       type = false;
       break;
   }
   const address = props.address !== props.senderId ? props.senderId : props.recipientId;
   const template = type ?
-    <span className={sytles.smallButton}>{type}</span> :
-    <ClickToSend recipient={address} className='from-to' >
+    <span className={styles.smallButton}>{t(type)}</span> :
```
slaweet: Calling `t(type)` doesn't allow i18n-scanner to find type names for the source *.json file.

src/utils/saveAccount.js
```diff
@@ -2,16 +2,17 @@ export const getSavedAccount = () => {
   const savedAccounts = localStorage.getItem('accounts');
   let account;
   if (savedAccounts) {
-    account = JSON.parse(savedAccounts)[0];
+    account = JSON.parse(savedAccounts);
```
slaweet: I don't see why `[0]` was removed.

reyraa: Since we aim to have multiple accounts there, I've implement the reducer to handle multiple. So I always should receive an array, even so currently the utility returns only one item. 
There are also other problems with this utility which I found it irrelevant to this PR so I didn't touch.

features/accountManagement.feature
```diff
@@ -1,6 +1,10 @@
 Feature: Account management
   Scenario: should allow to save account locally, after page reload it should require passphrase to do the first transaction, and remember the passphrase for next transactions
-    Given I'm logged in as "genesis"
+    Given I'm on login page
+    And I select option no. 3 from "network" select
+    When I fill in "http://localhost:4000" to "address" field
+    When I fill in "wagon stock borrow episode laundry kitten salute link globe zero feed marble" to "passphrase" field
+    And I click "login button"
```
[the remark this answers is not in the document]
reyraa: fixed

src/components/header/index.test.js
```diff
@@ -1,25 +1,31 @@
...
+    activePeerSet: () => {},
+  });
 
   beforeEach(() => {
-    wrapper = mount(<Provider store={store}>
+    wrapper = mount(<Provider store={store}><Router>
       <I18nextProvider i18n={ i18n }>
         <HeaderHOC />
-      </I18nextProvider>
+      </I18nextProvider></Router>
```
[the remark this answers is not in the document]
reyraa: fixed

src/utils/saveAccount.js
slaweet: then `account` should be `accounts`

src/components/tabs/tabs.js
reyraa: Done

src/components/transactions/transactionType.js
reyraa: Done

src/components/relativeLink/index.js
```diff
@@ -0,0 +1,24 @@
...
+  location, to, children, className, raised, neutral, primary, flat, disableWhenOffline,
+}) => {
+  let style = '';
+  if (raised !== undefined) style += `${buttonStyle.raised} `;
+  if (neutral !== undefined) style += `${buttonStyle.neutral} `;
+  if (flat !== undefined) style += `${buttonStyle.flat} `;
+  if (primary !== undefined) style += `${buttonStyle.primary} `;
+  if (disableWhenOffline !== undefined) style += `${offlineStyle.disableWhenOffline} `;
+  if (style !== '') style += ` ${buttonStyle.button}`;
+
```
alepop: I think that https://github.com/JedWatson/classnames is better solution for dynamic classnames

src/components/saveAccount/saveAccount.js
```diff
@@ -1,36 +1,47 @@
...
+  publicKey,
+  closeDialog,
+  accountSaved,
+}) => {
+  const save = () => {
+    // eslint-disable-next-line arrow-body-style
+    const index = networksRaw.map((item, i) => {
+      return (item.name === network) ? i : null;
+    }).filter(item => item !== null)[0];
+    accountSaved({
```
alepop: `.find(item => item ! == null);` will return first element that not equal null

src/store/middlewares/savedAccounts.js
```diff
@@ -0,0 +1,18 @@
+import actionTypes from '../../constants/actions';
+import { successToastDisplayed } from '../../actions/toaster';
+
+const savedAccountsMiddleware = store => next => (action) => {
+  next(action);
+  switch (action.type) {
+    case actionTypes.accountSaved:
+      store.dispatch(successToastDisplayed({ label: 'Account saved' }));
+      break;
```
alepop: why not use `return` instead of `break` in the switch case? It will look more compact.

src/utils/saveAccount.js
reyraa: There's a ticket to fix many other issues about this utility.

src/components/relativeLink/index.js
reyraa: There's a ticket to fix many other issues about this utility.

src/store/middlewares/savedAccounts.js
reyraa: There's a ticket to fix many other issues about this utility.

src/components/saveAccount/saveAccount.js
reyraa: Done

PR #754, opened 2017-09-18
Add selected votes counts bar - Closes #445

### LINKED ISSUE #445: [Feature request] Voting: Always visible total number of (un)votes and notification when reached 33 (un)votes
### Expected behaviour
A always visible number of (un)votes I've already clicked. Like, if I've removed 12 votes and added 3, it should always show me the total number (15) and be visible all the time, while scrolling. Maybe the whole table-header should be fixed visible.

It also should notify me, when I've reached 33 (un)votes and tell me that the number of votes for one transactions has reached the limit.

### Actual behaviour
You have to count the clicks for yourself, or scroll up and calculate the number of (un)votes by yourself.
You also can (un)vote as many delegates as you want, but after you click on vote only the info-text shows me, that I've currently too many (un)votes done. Users have to figure this out by themself, which isn't optimal. 

### Steps to reproduce
It's obvious.

slaweet:
The goal is to add a bar fixed at the bottom of the voting page that displays numbers of votes selected for voting.

I had it previously implemented in Angular and it looked like this:
votes-bar
The original PR was https://github.com/LiskHQ/lisk-nano/pull/447

Closes #445


src/components/voting/votingBar.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import grid from 'flexboxgrid/dist/flexboxgrid.css';
+
+import votingConst from '../../constants/voting';
+import style from './votingBar.css';
+
+const VotingBar = ({ votes }) => {
+  const { maxCountOfVotes, maxCountOfVotesInOneTurn } = votingConst;
```
reyraa: To prevent verbosity please perform this shorthand assignment while importing

src/components/voting/votingBar.js
```diff
@@ -0,0 +1,50 @@
...
+  const votedList = Object.keys(votes).filter(key => votes[key].confirmed);
```
reyraa: Please use a single `forEach` loop instead of 3 `filter`s and file the lists thought the same iteration

src/store/middlewares/voting.js
```diff
@@ -0,0 +1,23 @@
...
+import { errorToastDisplayed } from '../../actions/toaster';
+
+const votingMiddleware = store => next => (action) => {
+  next(action);
+  if (action.type === actionTypes.voteToggled) {
+    const { votes } = store.getState().voting;
+    const voteCount = Object.keys(votes).filter(
+      key => votes[key].confirmed !== votes[key].unconfirmed).length;
+    const currentVote = votes[action.data.username] || { unconfirmed: true, confirmed: false };
+    console.log(voteCount, votingConst.maxCountOfVotesInOneTurn, currentVote);
```
reyraa: Remove `console.log`. It's interesting that you have a log here and `eslint` doesn't fail

src/store/middlewares/voting.js
```diff
@@ -0,0 +1,23 @@
...
+    if (voteCount === votingConst.maxCountOfVotesInOneTurn + 1 &&
```
reyraa: Would be great if we show a toast for when we exceed 101 overall votes.

src/components/voting/votingBar.js
slaweet: I wanted to use it like that, but it doesn't work. 
```
import { maxCountOfVotes, maxCountOfVotesInOneTurn } from '../../constants/voting';
```
ends up in both `maxCountOfVotes` and `maxCountOfVotesInOneTurn` being `undefined`

src/components/voting/votingBar.js
slaweet: I set up a jsperf to see the difference:
https://jsperf.com/multiple-filters-vs-foreach

ForEach is 3 or 4 times faster, but the absolute values of how long it takes on a list of 100 still make the difference negligible.

I like using `fliter`s for their conciseness and readability.

PR #759, opened 2017-09-19
Find all strings for i18n - Closes #557

### LINKED ISSUE #557: Find all strings for i18n
This task is the second step of #3.
The goal is to go through the whole code base and update every string shown in UI is processed by the transaction function.

yasharAyari:
Closes [#557](https://github.com/LiskHQ/lisk-nano/issues/557)


src/components/forging/forgingTitle.js
```diff
@@ -21,7 +22,7 @@ class ForgingTitle extends React.Component {
               {this.props.account.delegate.username}
             </h2>
             <span>
-              <LiskAmount val={this.props.statistics.total} roundTo={2} /> LSK Earned
+              <LiskAmount val={this.props.statistics.total} roundTo={2} /> LSK {this.props.t('Earned')}
```
slaweet: `LSK` should be part of the translated string `{this.props.t('LSK Earned')` because other languages can have a different word order.

src/components/header/header.js
```diff
@@ -29,29 +30,29 @@ const Header = props => (
...
             },
             childComponent: SignMessage,
           })}
         />
-        <MenuItem caption="Verify message"
+        <MenuItem caption={props.t('Verify message')}
           className='verify-message'
           onClick={() => props.setActiveDialog({
-            title: 'Verify message',
+            title: props.t('verify-message'),
```
slaweet: This key should be the same as the one two lines ago - `'Verify message'`

src/components/header/header.js
```diff
@@ -64,6 +65,12 @@ const Header = props => (
           title: props.t('send'),
           childComponent: Send,
         })}>{props.t('send')}</Button>
+        <IconMenu
+          className={`${styles.iconButton} ${offlineStyle.disableWhenOffline}`}
+          icon='language' position='topRight'>
+          <MenuItem caption='english' onClick={() => i18n.changeLanguage('en')} />
```
slaweet: The first letter of caption should be capitalized `caption='English'`

src/components/header/header.js
```diff
@@ -64,6 +65,12 @@ const Header = props => (
...
+          <MenuItem caption='deutsch' onClick={() => i18n.changeLanguage('de')} />
```
slaweet: The first letter of caption should be capitalized `caption='Deutsch'`

src/locales/en/common.json
```diff
@@ -1,4 +1,49 @@
 {
   "send": "send",
```
slaweet: S in send should be capitalized: `"send": "Send",` It's the same on the button, but different as the send modal title.

src/components/forging/forgingStats.js
```diff
@@ -41,9 +42,13 @@ class ForgingStats extends React.Component {
               <CardText>
               <div className={grid['col-xs-12']}>
                 <div className={`${grid.row}  ${grid['between-xs']}`}>
-                  <span className='title'> {cardObj.label} </span>
+                  <span className='title'>
+                    { cardObj.label === 'Last 24 hours' ?
+                      this.props.t(cardObj.label) :
+                      this.props.t('Last x days', { day: cardObj.label }) }
```
slaweet: This condition is ugly and hacky. I would move the definition of `statCardObjects` inside `ForgingStats.render` so that it can use `this.props.t`and here the code can stay the way it was.

slaweet: The key should be `Last {{day}} days` for better readability. Or maybe `Last {{count}} days` as well as the value.

slaweet (changes requested):
Thanks, @yasharAyari .

Please see some comments below.

Also, there should be an e2e test for switching language.

src/components/header/header.js
```diff
@@ -64,6 +65,12 @@ const Header = props => (
           title: props.t('send'),
           childComponent: Send,
         })}>{props.t('send')}</Button>
+        <IconMenu
+          className={`${styles.iconButton} ${offlineStyle.disableWhenOffline}`}
+          icon='language' position='topRight'>
```
slaweet: The `IconMenu` can use `selectable` and `selected` properties to indicate selected language.
http://react-toolbox.com/#/components/menu

slaweet (dismissed):

slaweet (approved):

PR #760, opened 2017-09-19
Setup css-hot-reload - Closes #752

### LINKED ISSUE #752: CSS hot reload is broken
### Expected behaviour
When in running `npm run dev` and changing a css file, it should update the styles on the page by itself.
### Actual behaviour
When in running `npm run dev` and changing a css file, I have to refresh the page to see the new css.

alepop:
close #752


slaweet (approved):
Works like a charm. Thanks @alepop

PR #763, opened 2017-09-22
Find all strings for i18n in modals - Closes #744

### LINKED ISSUE #744: Find all strings for i18n in modals
This is a follow-up issue to #557, which turned to be bigger than expected, so this part will be done separately.

slaweet:
Closes #744


reyraa:
We have a navigate function which uses tabs arrays to navigate to tabs:
```
const navigate = (history, tabs, index) => {
  if (!isCurrent(history, index, tabs)) {
    history.push(`/main/${tabs[index].toLowerCase()}`);
  }
};
```
translating tabs items breaks the navigation since the route is not defined anymore. I think we have to use objects instead of string in this array:
```
const tabs = [
  {
    label: t('Transactions'),
    id: transactions,
  }, {
    label: t('Voting'),
    id: voting,
  }, {
    label: t('Forging'),
    id: forging,
 },
];
```

reyraa:
These translations are missing:
  - settings label in menu
  - all the labels in votingBar
  - The description in forging for non-delegates

PR #765, opened 2017-09-22
Setup auto update - Closes #196

### LINKED ISSUE #196: Notify users when there is new version available
### Expected behaviour
When a user launches Lisk Nano and it's not the latest version, some kind of unobtrusive notification should inform the user about the newer version. Possibly with a download link.
### Actual behaviour
When a new version of Lisk Nano is released, users have to follow Lisk communication channels (chat, blog, etc.) to find out about it or check Lisk website themeselves.

slaweet:
### What was the problem?
Users stay stayed on older versions without knowing there is a new one.

### How did I fix it?
By setting up [electron-updater](https://www.electron.build/auto-update)

### How to test it?
- decrease version in `app/package.json` to `1.2.0`
- build nano with `npm run pack` (On mac) or `npm run pack:win` (On Windows)
- install the built package from 'dist' folder
- open lisk nano
- wait for couple seconds
- there will be an alert like this: screen shot 2017-11-17 at 14 30 13
- click "Update now"
- wait for a minute or two for the download (you can use that time to check version in about dialog or log in and check that "encrypt message" feature is present)
- there will be another alert when the download finishes, confirm it.
- the app restarts
- check version in about dialog or log in and check that "encrypt message" feature is no longer present

### Review checklist
- [ ] The PR solves #196
- [ ] All new code is covered with unit tests
- [ ] All new code follows best practices
- [ ] Works on Mac and Windows (Linux will be handled separately https://github.com/LiskHQ/lisk-nano/issues/991)


ginacontrino:
@slaweet when I click on "update", it takes like 30 seconds or more before I get the notification "Updates downloaded, application will be quit for update...".  This was quite confusing, because I was wondering if it's updating or not. Usually you get some kind of progress bar or some other kind of notification that something is actually happening.

Also, when I download the app, and I move it to the "Applications", but open it from here:

screen shot 2017-11-17 at 15 10 53

And then click "update", it is not actually updating. I, as a user, often open the app directly from that window, so I can imagine other people doing that as well.

PR #770, opened 2017-09-25
Setup browser language detector - Closes #556

### LINKED ISSUE #556: Setup internationalization framework
This task is the first step of #3.
It includes:
- Find and choose an internationalization library (there are many e.g. https://github.com/i18next/react-i18next https://github.com/yahoo/react-intl ) 
- Set it up to read the default language from OS via Electron.
- Add a setting option for users to change the language

slaweet:
This is the last part left from #556


slaweet:
I think that the "automatic translations" is valid with websites, but Nano is a desktop app you're unlikely to install on another machine then your own. 

I will look into a way to resolve locales like `en-US` and `en-UK` to `en`. Similarly for `de`.

reyraa:
I still find it annoying even in mobile apps (which run in the most personal medium), but If other teammates consent, I'm down with it.

slaweet:
As for locales resolving (e.g. `en-US` to `en`) it works out of the box.
https://www.i18next.com/principles/fallback.html#locals-resolving
It first tries `en-US` and if not found goes to `en`

slaweet:
I gave it some thought and in the end implemented a language switcher dropdown on login page.
I believe it solves the problem of autodetect selecting a language that the user doesn't want. 
At the same time, it doesn't add any extra step/clicks of choosing/confirming the language in most cases when the autodetect chooses what user wants.

reyraa:
Thanks Vit

Issue #736, opened 2017-09-12
Choose and set up a localization platform


slaweet:
After i18n is set up in https://github.com/LiskHQ/lisk-nano/issues/3 we should set up a localization platform to separate translation management from development.

With a localization platform, we can store the translations separately from the source code and download the latest translations automatically as one step of `npm run build`. 

There should also be a command for downloading the latest translations when necessary. 

Uploading latest transaction sources can be part of building `development` branch in Jenkins.

I have personally good experience with https://www.transifex.com
But there are other options, like https://crowdin.com

slaweet:
Done. We use https://translate.lingohub.com/liskhq/lisk-nano/dashboard

PR #771, opened 2017-09-25
Setup a i18n scanner command - Closes #769

### LINKED ISSUE #769: Setup a i18n scanner command
### Expected behaviour
There should be a command to run that generates the `*.json` translation source file, based on occurances of `this.props.t(`, `props.t(`, `t(`, and `i18next.t(`
e. g. https://github.com/i18next/i18next-scanner/wiki/Integration-Guide#react
### Actual behaviour
There's none

slaweet:
Closes #769


src/i18n-scanner.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs');
+const glob = require('glob');
+const Parser = require('i18next-scanner').Parser;
+
+const translationFunctionNames = ['i18next.t', 'props.t', 'this.props.t', 't'];
+const outputFilePath = './src/locales/en/common.json';
```
reyraa: Are we supposed to update the json file of other languages manually? I don't mean the values, but the keys.

reyraa:
How it keeps track of the changed/removed translations? say, If for any reason we change a title's casing (e.g from transaction to Transaction), will we have to search git history for the missing translation?

src/locales/en/common.json
```diff
@@ -1,58 +1,57 @@
...
+  "confirmations": "confirmations",
   "my votes": "my votes",
-  "Search": "Search",
-  "Name": "Name",
-  "Lisk Address": "Lisk Address",
-  "Uptime": "Uptime",
-  "No delegates found": "No delegates found"
-}
+  "your second passphrase will be required for all transactions sent from this account": "your second passphrase will be required for all transactions sent from this account"
+}
```
reyraa: I guess we need an ending line here

src/i18n-scanner.js
slaweet: Other translation files are supposed to be changed by the localization platform set up in https://github.com/LiskHQ/lisk-nano/issues/736

In the end, I'd prefer to .gitignore the translation files and fetching them from the localization platform on a build.

reyraa:
So the remaining points to be cleared are:
 - How does it handle different translations in different contexts
 - How does it handle plural/singular translations

slaweet:
context and plurals is something that `i18n-scanner` can take care of based on `context` and `count` options, e.g. `props.t('{{count}} confirmations', {count: 2, context: 'tooltip'})` will by default create keys `{{count}} confirmations_tooltip`, `{{count}} confirmations_tooltip_plural`.

But I broke it by using `customHandler` to get the values prepopulated:
https://github.com/LiskHQ/lisk-nano/blob/17b192ed9334d831aa4937a77aa4ca40f28fedf7/src/i18n-scanner.js#L10-L12

I don't know how to get values prepopulated at the same time as context and plurals

slaweet:
I gave it another try and fixed the plurals and context

PR #778, opened 2017-09-26
Cleanup Jenkins stages

slaweet:
Merge short ones and update labels when necessary


fchavant:
We should remove the explicit calls to`#!/bin/bash`: they prevent the `sh` steps from failing if anything but the command or their last line fails.

PR #792, opened 2017-09-29
Prevent NPE for recovery notification

ghost:
- check if there is a previous build before trying to find out its result
- add PR branch name to the recovery notification


slaweet (approved):
Thanks @fchavant

PR #845, opened 2017-10-09
Parallel jenkins builds - Closes #825

### LINKED ISSUE #825: Run Jenkins builds in paralel
### Expected behaviour
Jenkins could do multiple builds in parallel so that we have to wait less.
We just need to make sure that the builds have don't crash each other, e.g. use distinct ports, separate Lisk Core instance. 

### Actual behaviour
Currently, only one build at the time can be run in Jenkins. Sometimes we have to wait for several unrelated builds to finish to have one done.

ghost:


slaweet (dismissed):
This is great. Thank you @fchavant

slaweet (approved):

PR #848, opened 2017-10-10
Adding more tests for register component - Closes #839 

### LINKED ISSUE #839: Improve unit test coverage of register component
### Expected behaviour
The test coverage of [register component](https://github.com/LiskHQ/lisk-nano/tree/development/src/components/register) should be close to 100%
### Actual behaviour
The test coverage is poor in register component
https://coveralls.io/builds/13570932/source?filename=src%2Fcomponents%2Fregister%2Fregister.js

ginacontrino:
What's the Problem?
The test coverage of the register component was poor + the component needed some refactoring [#839](https://github.com/LiskHQ/lisk-nano/issues/839)

What did you do?
- Added a login util to get rid of duplicated code
- Tested the utils file
- Added tests for the register component

How can I test this?
- Run tests
- Also check in the browser that the login and registration still work as expected

screen shot 2017-10-10 at 09 07 43

### DIFF
--- a/src/components/register/register.test.js
+++ b/src/components/register/register.test.js
@@ -13,6 +13,7 @@ import * as Utils from '../../utils/login';
 
 describe('Register', () => {
   let wrapper;
+  let loginData;
   const peers = { data: {} };
   const account = {};
   const store = configureMockStore([])({
@@ -36,6 +37,11 @@ describe('Register', () => {
 
   beforeEach(() => {
     wrapper = mount(<Register {...prop} />, options);
+    loginData = stub(Utils, 'getLoginData');
+  });
+
+  afterEach(() => {
+    loginData.restore();
   });
 
   it('renders Passphrase component', () => {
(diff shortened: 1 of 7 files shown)


src/components/register/register.test.js
```diff
@@ -56,4 +59,54 @@ describe('Register', () => {
       passphrase: 'sample passphrase',
     });
   });
+
+  it('should call activePeerSet with testnet if network index is testnet', () => {
+    const props = wrapper.find('Passphrase').props();
+
+    const loginData = stub(Utils, 'getLoginData');
```
slaweet: Please move the stub creation to `beforeEach` block and the stub `restore` to `afterEach` block. It avoids some code duplication and makes tests more stable because `restore` gets called even if there is some error in the middle of the test and no other test is affected. WIth `restore` inside the test case, in case of error in the test, the stub is not restored and the next test fails too because it is trying to stub something that already is a stub.

PR #854, opened 2017-10-11
Deploy after build stage and notify GitHub - Closes #853

### LINKED ISSUE #853: Perform test deployments after build and report status to GitHub
### Expected behaviour

Jenkins test deployments should be performed right after the "Build Nano" stage; deployment status should be reported to GitHub as a commit status.

### Actual behaviour

Jenkins test deployments are performed only after end-to-end tests run successfully; deployment status is not reported to GitHub.

### Steps to reproduce

- look at a PR build in Jenkins

ghost:
Move test deployment so that it happens right after the "Build nano"
stage and notify GitHub status API with custom context.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -74,6 +74,16 @@ node('lisk-nano') {
       }
     }
 
+    stage ('Deploy') {
+      try {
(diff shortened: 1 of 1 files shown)


slaweet (approved):
Great. Thanks @fchavant

PR #865, opened 2017-10-13
Cache node_modules - Closes #864

### LINKED ISSUE #864: Cache node_modules
### Expected behaviour

The nightly build of lisk-nano caches the installed node_modules so that other builds do not have to download and install everything from scratch.

### Actual behaviour

Every build starts with an empty node_modules directory and downloads every required package.

### Steps to reproduce

Run any lisk-nano build on Jenkins, observe npm downloading and installing every single package and its dependencies.

ghost:
Cache `node_modules/` directory from nightly builds (development) and use cache to speed up all builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,7 +44,14 @@ node('lisk-nano') {
 
     stage ('Install npm dependencies') {
       try {
-        sh 'npm install'
+        sh '''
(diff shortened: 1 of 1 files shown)


slaweet (approved):
Wow, that's fast. Thanks @fchavant

PR #869, opened 2017-10-14
Add support for voting / unvoting any amount of delegates

### LINKED ISSUE #868: Allow for selection of < 33 delegates to vote / unvote.
Lisk allows for only 33 delegates to be voted / unvoted for per transaction, but there is no reason not to allow people to select > 33 delegates and inform them that they will pay an additional fee.

This is a lot user friendlier if you need to vote in bulk.

Removed-5an:
closes https://github.com/LiskHQ/lisk-nano/issues/868

Looking for feedback especially on how to handle errors and will need some help with writing the tests.


slaweet:
I think the idea of this PR is a very smart way to work around the limit in Lisk Core. A first I was very excited by that.

But later I noticed that the PR doesn't cover several things:
- error handling: what happens if for whatever reason (e.g. bad internet connection) some of the requests succeed and some fail.
- if a user does votes and unvotes and is close to the total limit of 101 votes, it does the votes before unvotes, which may cause "over the 101 limit" error.
- doesn't cover end-to-end tests 
- doesn't update unit tests.

Especially with the error handling, I'm not sure there is a good way to handle it properly without making the voting a lot more complicated. If I were to consider merging this, first I would like to see a how the error handling should be handled.

5an1ty:
As discussed on gitter:

1. unvotes will now be executed first
2. vote transactions will happen in series, not parallel
3. Added some initial error handling, an error message will be shown if any of the transactions fails and only those that succeeded will move to pending, the others will stay in unvote, vote state so you can easily click to try submitting again.

slaweet:
Hi Ruben, we’ve discussed it thoroughly with Oliver and came to the conclusion that we don't want to accept the change. The reason is that it introduces more complexity to the voting (mainly for developers, but also for users). Voting is already the most complex part that attracts bugs and we don't want to complicate it even more. While we see the benefit of your changes, we don't think it is worth the increased complexity. We will proceed with the proper (but slower) way to address the limit, which is changing it in Lisk Core.

Issue #942, opened 2017-11-02
Lisk Nano throw an error if config.json is corrupted


alepop:
### Expected behaviour
If file in `appData` folder is corrupted (`storage/config.json`) it causes an error on Lisk Nano launch.
It happens because `electron-json-storage` expects file with valid json format, but file can be empty.

![screen shot 2017-11-02 at 10 42 46](https://user-images.githubusercontent.com/922338/32315394-f3de06e4-bfbc-11e7-9401-d7bd787a8285.png)

### Actual behaviour
There is no error exception on Lisk Nano launch.

### Steps to reproduce
Try to open and close Lisk Nano many times before full loading is completed.

slaweet:
Fixed in #943

PR #943, opened 2017-11-02
Fix exception throwing - Close 942

alepop:
### What was the problem?
If file in `appData` folder is corrupted (`storage/config.json`) it causes an error on `Lisk Nano` launch.
It happens because `electron-json-storage` expects file with valid `json` format, but file can be empty.

### How did I fix it?
Remove `throw error` code.

### How to test it?
Try to open and close Lisk Nano many times before full loading is completed.

### Review checklist
- [x] The PR solves #942
- [x] All new features are covered with e2e tests


slaweet (approved):
Thank you @alepop

PR #950, opened 2017-11-03
Fix recovery notification - Closes #944

### LINKED ISSUE #944: Build recovery slack notifications don’t work
### Expected behaviour
There is code that should send a message to our slack channel if build passed and the previous build failed.
https://github.com/LiskHQ/lisk-nano/blob/f0d37cbd063a152f424b6217f986d47f8d94c8f7/Jenkinsfile#L162-L167
### Actual behaviour
It doesn't send any recovery notifications since https://github.com/LiskHQ/lisk-nano/issues/825

ghost:
It appears currentBuild.result can be null if the build is passing.
Ref. https://jenkins.io/doc/book/pipeline/jenkinsfile/#deploy

### What was the problem?

It appears `currentBuild.result` can be `null` for a successful build.

### How did I fix it?

Changed the success condition to check not only for `"SUCCESS"` but also for `null`.

### How to test it?

- make the build fail (e.g. by using Jenkins' replay feature executing `sh 'false'`)
  - a failure notification should be sent to Slack
- build again normally
  - this time the build should not fail
  - a recovery notification should be sent to Slack

### Review checklist
- [x] The PR solves #944
- [x] All new features are covered with e2e tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,8 @@ def fail(reason) {
   error("${reason}")
 }
 
+/* comment out the next line to allow concurrent builds on the same branch */
+properties([disableConcurrentBuilds(), pipelineTriggers([])])
 node('lisk-nano') {
   try {
     stage ('Cleanup, Checkout and Start Lisk Core') {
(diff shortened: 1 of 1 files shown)


slaweet (approved):
It works :-) Thank you @fchavant

PR #960, opened 2017-11-06
Setup testnet tests - Closes #925

### LINKED ISSUE #925: E2e test send on testnet
There should be some e2e tests that run with testnet:
- send with testnet
- vote with testnet
- send with Custom Node https://testnet.lisk.io
- vote with Custom node https://testnet.lisk.io

There should be a way to specify a testnet passphrase as an argument of e2e test command, e.g.:
`npm run --silent e2e-test -- --params.testenetPassphrase 'wagon stock borrow episode laundry kitten salute link globe zero feed marble'` so that we don't need to have the passphrase in the public source code. In Jenkins we can have the passphrase in an environment variable.

The default of the `testenetPassphrase` param should be something like `null`, defined here:
https://github.com/LiskHQ/lisk-nano/blob/d69cc1c2f6baf918c9eaea13bb0f41e6c3b8325c/protractor.conf.js#L23-L27

If the `testenetPassphrase` param is `null`, then the step can be marked as pending, using: `callback(null, 'pending');` [0], so that it is possible to successfully run the e2e tests without a testnet account.

[0] https://github.com/cucumber/cucumber-js/issues/873#issuecomment-325648911

ginacontrino:
### What was the problem?
We had no configuration for running tests with testnet

### How did I fix it?
- Added scripts so we can run some of the existing voting and sending tests with testnet 

### How to test it?
Before you run the tests, you should save the passphrase in your env
`export TESTNET_PASSPHRASE='recipe bomb asset salon coil symbol tiger engine assist pact pumpkin visit'`
- Then, in order to just run tests with testnet you can go:
`npm run e2e-test -- --disableChecks --cucumberOpts.tags=@testnet --params.network=testnet --params.useTestnetPassphrase=true`
- In order to run tests with the custom node "https://testnet.lisk.io" you can go:
`npm run e2e-test -- --disableChecks --cucumberOpts.tags=@testnet -- --params.liskCoreURL=https://testnet.lisk.io --params.useTestnetPassphrase=true`

- Make also sure that the normal tests still work:
`npm run e2e-test`

### Review checklist
- [ ] The PR solves https://github.com/LiskHQ/lisk-nano/issues/925
- [ ] All new code follows best practices

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -120,15 +120,15 @@ node('lisk-nano') {
         ansiColor('xterm') {
           sh '''
           N=${EXECUTOR_NUMBER:-0}
-          passphrase=${env.TESTNET_PASSPHRASE}
 
           # End to End test configuration
           export DISPLAY=:1$N
           Xvfb :1$N -ac -screen 0 1280x1024x24 &
 
           # Run end-to-end tests
-          npm run --silent e2e-test:testnet:custom -- --params.baseURL file://$WORKSPACE/app/build/index.html --params.liskCoreURL http://127.0.0.1:400$N  --params.testnetPassphrase $passphrase --params.testnetCustomNode true
-          npm run --silent e2e-test:testnet -- --params.baseURL file://$WORKSPACE/app/build/index.html --params.liskCoreURL http://127.0.0.1:400$N --params.testnetPassphrase $passphrase
+
+          npm run --silent e2e-test:testnet:custom -- --params.baseURL file://$WORKSPACE/app/build/index.html --params.liskCoreURL http://127.0.0.1:400$N --params.testnetCustomNode true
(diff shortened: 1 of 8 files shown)


Jenkinsfile
```diff
@@ -126,6 +126,14 @@ node('lisk-nano') {
           Xvfb :1$N -ac -screen 0 1280x1024x24 &
 
           # Run end-to-end tests
+
+          withCredentials([string(credentialsId: 'lisk-nano-testnet-passphrase', variable: 'TESTNET_PASSPHRASE')]) {
```
fchavant: You need to move this `withCredentials` statement to wrap the whole `sh` above, i.e. you need to have:
```[...]
withCredentials([...]) {
    sh ```
        [...]
        npm run --silent e2e-test:testnet:custom -- --params.baseURL file://$WORKSPACE/app/build/index.html --params.liskCoreURL http://127.0.0.1:400$N  --params.testnetPassphrase $TESTNET_PASSPHRASE --params.testnetCustomNode true
        [...]
    ```
}
[...]```

PR #976, opened 2017-11-10
Use core port number > 4000 to prevent side effects

ghost:
### What was the problem?

E2E tests would sometimes use the default core port although they're supposed to be completely isolated. 

### How did I fix it?

Use port numbers > 4000

### How to test it?

Watch jenkins builds, make sure integration tests do not fail any more often than they used to.


slaweet (approved):
Looks fine.

PR #1000, opened 2017-11-22
Fix autocomplete when changing a vote - Closes #990

### LINKED ISSUE #990: Autocomplete changing a vote returns no matches for a valid delegate
### Expected behaviour
On the "Vote For Delegates" modal, when you search for a valid delegate, add a vote to them, and then remove it, they no longer show up in the autocomplete. An error is returned stating "No matches". Refreshing the dialog doesn't fix this and a full app refresh is needed to return the autocomplete back to the intended state.

### Actual behaviour
The delegate should appear in the autocomplete after they've been removed from your voting selection. 

### Steps to reproduce
- Go to the "Voting" tab
- Click "Vote"
- Search for a valid delegate(ie. "kippers")
- Select the delegate from the autocomplete list
- Click the "x" button next to the delegate
- Search for the delegate again
- An error will be returned stating "No matches"

pouljohn1:
### What was the problem?

Described here #990 

### How did I fix it?

Previously when we had toggled and untoggled specified vote, reducer state was with false state (vote.confirmed and vote.unconfirmed were false). I've changed it in a way that when user untoggled vote and state of this vote is unconfirmed I'm removing this vote from state, so final state is the same as initial state (before toggling vote).

### How to test it?

Described here #990 

### Review checklist
- [x] The PR solves #990 

### DIFF
--- a/src/store/reducers/voting.js
+++ b/src/store/reducers/voting.js
@@ -66,20 +66,26 @@ const voting = (state = {
         refresh: true,
       });
 
-    case actionTypes.voteToggled:
+    case actionTypes.voteToggled: {
+      const { username, publicKey } = action.data;
+
+      const newVotes = Object.assign({}, state.votes);
+      const currentVote = state.votes[username];
+      if (currentVote && currentVote.unconfirmed && !currentVote.confirmed) {
+        delete newVotes[action.data.username];
+      } else {
+        newVotes[username] = {
+          confirmed: currentVote ? currentVote.confirmed : false,
+          unconfirmed: currentVote ? !currentVote.unconfirmed : true,
+          publicKey,
+        };
+      }
+
       return Object.assign({}, state, {
         refresh: false,
-        votes: Object.assign({}, state.votes, {
-          [action.data.username]: {
-            confirmed: state.votes[action.data.username] ?
-              state.votes[action.data.username].confirmed : false,
-            unconfirmed: state.votes[action.data.username] ?
-              !state.votes[action.data.username].unconfirmed : true,
-            publicKey: action.data.publicKey,
-          },
-        }),
+        votes: newVotes,
       });
-
+    }
 
     case actionTypes.accountLoggedOut:
       return Object.assign({}, state, {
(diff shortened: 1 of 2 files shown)


reyraa:
We appreciate your participation. Please consider the following tweak. in `src/utils/api/delegate.js#L26`, apply this change:
```
return new Promise((resolve, reject) =>
    listDelegates(activePeer, options)
      .then((response) => {
        resolve(response.delegates.filter(delegate =>
          Object.keys(votedDict).filter(item =>
            (item === delegate.username) && (item.confirmed || item.unconfirmed)).length === 0,
        ));
      })
      .catch(reject),
  );
```
I've added `item.confirmed || item.unconfirmed` to the condition and it fixes the issue.

paweljaneczek:
It was my first attempt, but for me it is more like hack. As I understand state like !unconfirmed and !confirmed should not happen (what does it mean than?) - for me it should be either confirmed or unconfirmed. 
My approach results the same state as it was before toggling vote on which I think is a good approach.

But of course I don’t know whole code so maybe missing something. If you want I could change to this.

slaweet:
I see that the usage of `unconfirmed` and  `confirmed` is not very clear. 
 - `unconfirmed` says if the delegate is selected or not in the app.
 - `confirmed` says if the delegate is selected or not on the blockchain.

`!unconfirmed and !confirmed` means that the delegate selected neither in the app nor on the blockchain.

As @reyraa mentioned the bug can be fixed with much fewer changes in this function:
https://github.com/LiskHQ/lisk-nano/blob/32cc92608f5c6c72dacf1f286d267ac2111a7943/src/utils/api/delegate.js#L26-L35

paweljaneczek:
I get this, but still I think that such vote should not be in reducer (it means the same). For me this reducer needs some refactor, it is too much error prone. 
But for not making too much changes I will do the way that you told

slaweet:
The proposed solution was implemented in https://github.com/LiskHQ/lisk-nano/pull/1074 so closing this one.

PR #1049, opened 2018-04-07
Fix random index calculation in hideRandomWord()

webmaster128:
This is a port of https://github.com/LiskHQ/lisk-hub/pull/652


reyraa (approved):
Thank you Simon.

PR #1068, opened 2018-05-15
Remove error toast after autoupdate - Closes #1062

### LINKED ISSUE #1062: Remove error toast after auto-update and auto-restart
### Expected behaviour
There should be no error toast after auto-update and auto-restart

### Actual behaviour
There is an error toast "URL invalid"

slaweet:
### What was the problem?
See #1062

### How did I fix it?
Removed the error toast

### How to test it?
That's close to impossible because you need the change to be released.

### Review checklist
- The PR solves #1062
- All new code is covered with unit tests
- All new features are covered with e2e tests
- All new code follows best practices

### DIFF
--- a/src/utils/externalLinks.js
+++ b/src/utils/externalLinks.js
@@ -1,8 +1,5 @@
-import i18next from 'i18next';
 import history from '../history';
 import routesReg from './routes';
-import { errorToastDisplayed } from '../actions/toaster';
-import store from '../store';
 
 export default {
   init: () => {
@@ -14,8 +11,6 @@ export default {
         const route = routesReg.find(item => item.regex.test(normalizedUrl));
         if (route !== undefined) {
           history.push(normalizedUrl);
-        } else {
-          store.dispatch(errorToastDisplayed({ label: i18next.t('The URL was invalid') }));
         }
       });
     }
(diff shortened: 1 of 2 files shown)


src/utils/externalLinks.js
```diff
@@ -1,21 +1,16 @@
...
-import { errorToastDisplayed } from '../actions/toaster';
-import store from '../store';
 
 export default {
   init: () => {
     const { ipc } = window;
 
     if (ipc) {
       ipc.on('openUrl', (action, url) => {
-        const normalizedUrl = url.toLowerCase().replace('lisk://', '/');
```
reyraa: If you click on `Lisk://main/transactions` this won't find the route. any IM may correct the URL from `lisk://` to `Lisk://`

src/utils/externalLinks.js
```diff
@@ -1,21 +1,16 @@
...
+        const normalizedUrl = url.replace('lisk://', '/');
         const route = routesReg.find(item => item.regex.test(normalizedUrl));
         if (route !== undefined) {
           history.push(normalizedUrl);
-        } else {
-          store.dispatch(errorToastDisplayed({ label: i18next.t('The URL was invalid') }));
```
reyraa: Please be aware that removing this line, we won't be notifying the user about probable mistakes in URL. for example, `lisk://mai/transactions` or `lisk://main/transaction` will result in no action and can be confusing.

PR #1077, opened 2018-05-25
Test against lisk core 1.0.0-beta.9.2

ghost:
### What was the problem?

Some tests need to run against lisk core >=1.0.0-beta.7

### How did I fix it?

Bumped version number.

### How to test it?

Run Jenkins build.


fchavant:
@slaweet there is only one test failing, I suspect the test itself to be problematic; can you have a look?

slaweet (approved):
Thank you @fchavant

